PR #12, opened 2012-04-26
option to skip html escaping

BurntSushi:
I found this option useful when allowing user submitted data to use Markdown. Namely, *all* of their data must be sanitized. However, blackfriday then escapes quotes and angle brackets inside code blocks and other various things, which makes the appearance less the palatable. (i.e., '&amp;' instead of '&' or '&quot;' instead of ".)

So I've added an option, HTML_SKIP_ESCAPE, that completely skips over all escaping when enabled.

### DIFF
--- a/block.go
+++ b/block.go
@@ -831,6 +831,12 @@ func (p *parser) quotePrefix(data []byte) int {
 		}
 		return i + 1
 	}
+	if len(data) > 4 && string(data[i:i+4]) == "&gt;" {
+		if data[i+5] == ' ' {
+			return i + 5
+		}
+		return i + 4
+	}
 	return 0
 }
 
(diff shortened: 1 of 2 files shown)


shurcooL:
Hey @BurntSushi,

This PR is from 2012. A lot has changed since then.

> However, blackfriday then escapes quotes and angle brackets inside code blocks and other various things, which makes the appearance less the palatable. (i.e., '&' instead of '&' or '"' instead of ".)

There is now an [`HTML_USE_SMARTYPANTS`](https://godoc.org/github.com/russross/blackfriday#HTML_USE_SMARTYPANTS) flag, which I think addresses that.

Can you see if the original motivation still applies? If so, this PR will need to be updated because there are merge conflicts. Otherwise, it can be closed.

BurntSushi:
Looks like `HTML_USE_SMARTYPANTS` should do the trick. Thanks!

Issue #90, opened 2014-06-13
Consider other days of the week?


shurcooL:
I've recently stumbled upon https://github.com/microcosm-cc/bluemonday, which seems to be a Go library for HTML sanitizing.

Would it be a good idea or a bad idea to use it?

I haven't really looked at it closely yet, but I just wanted to start the discussion here.

shurcooL:
This has been resolved in #117.

PR #117, opened 2014-09-19
Remove sanitization

rtfb:
Following discussion in #90, this seems to be the minimal first cut at what needs to be done to transfer sanitization duties to bluemonday: removed what we had in blackfriday and added a reference to the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ import (
 
 // ...
 unsafe := blackfriday.MarkdownCommon(input)
-html := bluemonday.UGCPolicy().Sanitize(string(unsafe))
+html := bluemonday.UGCPolicy().SanitizeBytes(unsafe)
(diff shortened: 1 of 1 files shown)


tianon:
It appears that this was the last place within blackfriday that was still referencing the now-dead `code.google.com/p/go.net/html` -- any chance we could get a new release tag to match so the latest release version builds properly without patches? :smile:  (cc @russross)

See https://github.com/docker/docker/issues/19787 for a little context.

tianon:
Wow, I'm special -- this PR was included in both 1.3 and 1.4; sorry for the noise!  Hopefully I can track down the correct PR without too much trouble...

tianon:
That Docker spot is still using v1.2 -- doh, no wonder it fails.

PR #134, opened 2014-12-16
Add programmable reference overrides

jtolds:
If a user provides a ReferenceOverride function, then reference ids
will be passed to the given ReferenceOverride function first, before
consulting the generated reference table.

The goal here is to enable programmable support for
"WikiWords"-style identifiers or other application-specific
user-generated keywords.

Example, writing documentation:

 The [Frobnosticator][] is a very important class in our codebase.
 While it is used to frobnosticate widgets in general, it can also
 be passed to the [WeeDoodler][] to interesting effect.

This might be solveable with the HTML Renderer relative prefix, but
I didn't see a good way of making a short link to 'Frobnosticator'
relatively without having to write it twice. Maybe
'<Frobnosticator>' should work? Should Autolinks work for relative
links?

In addition, I wanted a little more richness. I plan to support
Godoc links by prefixing references with a '!', like so:

  Check out the [Frobnosticator][] helper function
  [!util.Frobnosticate()][]

The first link links to the Frobnosticator architectural overview
documentation, whereas the second links to Godoc.

Better advice on how to implement this sort of think with
Blackfriday is highly desired.

### DIFF
--- a/inline.go
+++ b/inline.go
@@ -384,9 +384,8 @@ func link(p *parser, out *bytes.Buffer, data []byte, offset int) int {
 			id = data[linkB:linkE]
 		}
 
-		// find the reference with matching id (ids are case-insensitive)
-		key := string(bytes.ToLower(id))
-		lr, ok := p.refs[key]
+		// find the reference with matching id
+		lr, ok := p.getRef(string(id))
 		if !ok {
 			return 0
 
@@ -423,7 +422,6 @@ func link(p *parser, out *bytes.Buffer, data []byte, offset int) int {
 			}
 		}
 
-		key := string(bytes.ToLower(id))
 		if t == linkInlineFootnote {
 			// create a new reference
 			noteId = len(p.notes) + 1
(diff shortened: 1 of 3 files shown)


rtfb:
Hi @jtolds, thanks for your PR.

I love the idea and the implementation is neat. I'd like to merge it, but I have a couple notes regarding 8e10236: looks like this addition requires some more tests in `TestReferenceOverride` and I'd like the `snake_case` identifiers changed to `mixedCase`.

And never say no rush or it will end up hanging for several months like it did ;-)

@shurcooL, do you have anything to add?

shurcooL:
> @shurcooL, do you have anything to add?

No, nothing extra. My only comment was also to change `snake_case` identifiers to `mixedCase`.

jtolds:
Oh geez, sorry, I was totally going to fix the snake_case thing I just didn't get to it yet. Want me to still fix?

rtfb:
No problem, I already did. Please review commits 62f0018 and 314ce8f while you're at it.

jtolds:
oh great those look fantastic

PR #145, opened 2015-02-09
Move the license info from README to LICENSE file

eparis:
It's pretty standard to have a license file, and some projects require
them before your code can be used in their code.  Move the license to
its own file.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,0 +1,29 @@
+Blackfriday is distributed under the Simplified BSD License:
+
+> Copyright © 2011 Russ Ross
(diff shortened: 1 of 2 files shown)


rtfb:
In general I think the change is for the good, but I'd suggest leaving a minimalistic "License" section in the README with a reference to LICENSE.txt. Possibly with a mention that it's 2-clause BSD, so that people could see it at the first glance.

eparis:
Done.  With a link to the new file.

rtfb:
Cool, thanks!

PR #150, opened 2015-02-12
Update html.go

elian0211:
* update  func isRelativeLink .....   if a tag begin with '.' yes = true 
such as 
href="./"
href="../"

### DIFF
--- a/block.go
+++ b/block.go
@@ -556,8 +556,11 @@ func (p *parser) isFencedCode(data []byte, syntax **string, oldmarker string) (s
 	skip = 0
 
 	// skip up to three spaces
-	for i < 3 && data[i] == ' ' {
+	for i < len(data) && i < 3 && data[i] == ' ' {
 		i++
+	}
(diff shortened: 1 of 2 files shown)


rtfb:
Would be great if you could add tests while you're at it. There seem to be no tests for relative links as of now.

html.go
```diff
@@ -859,8 +859,8 @@ func doubleSpace(out *bytes.Buffer) {
 func isRelativeLink(link []byte) (yes bool) {
 	yes = false
 
-	// a tag begin with '#'
-	if link[0] == '#' {
+	// a tag begin with '#' or '.'
+	if link[0] == '#' || link[0] == '.' {
 		yes = true
 	}
```
shurcooL: This func `isRelativeLink` has 3 separate cases that may result in setting `yes` to `true`. Why did you decide to group '.' together with the first test case that checks if tag begins with '#', instead of adding a separate test case? It doesn't seem they are related.

PR #155, opened 2015-03-14
add rel="noreferrer" option

beyang:
This adds a flag to add a `rel="noreferrer"` attribute to generated links. This is generally a good idea when rendering third-party HTML with the `HTML_HREF_TARGET_BLANK` flag. Otherwise the linked-to page can use `window.opener.location.replace('http://attacker.com')` to replace the URL of the original tab, which is a potential security vulnerability.

Tests have been added for this case. Thanks for making this awesome library!

### DIFF
--- a/html.go
+++ b/html.go
@@ -31,6 +31,7 @@ const (
 	HTML_SKIP_LINKS                            // skip all links
 	HTML_SAFELINK                              // only link to trusted protocols
 	HTML_NOFOLLOW_LINKS                        // only link with rel="nofollow"
+	HTML_NOREFERRER_LINKS                      // only link with rel="noreferrer"
 	HTML_HREF_TARGET_BLANK                     // add a blank target
 	HTML_TOC                                   // generate a table of contents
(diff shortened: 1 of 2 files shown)


shurcooL:
This seems like a reasonable addition.

The code LGTM, thanks for adding tests!

robinpuri:
Fixed Nicely!

rtfb:
Thanks!

robinpuri:
is this fix live now?

beyang:
Yes it should be

robinpuri:
I don't think so! I was still able to reproduce this.

Also, Could you please check an xss issue sent to security@sourcegraph

robinpuri:
Hi, I don't think this is fixed and was also exploitable at certain endpoints.

beyang:
Hey @robinpuri, it's fixed in this library. Let's continue discussion over email to avoid spamming contributors here.

robinpuri:
sure!

Issue #158, opened 2015-03-26
Sequential note refs doesn't work


bep:
This works fine:

```
Some text.[^note1]
```
This renders only the second footnote correctly, the first is rendered as plain text.

```
Some text.[^note1][^note2]
```

PR #159, opened 2015-04-03
Fix recognition of consecutive footnotes

rtfb:
The second footnote was treated as if the pair of them were a reference
style link, without checking if the second bit is another footnote.

Fixes issue #158.


shurcooL:
LGTM.

PR #162, opened 2015-04-22
Implement backslash hard line break

neclepsio:
Many don't like double space because it's not visible. For this, CommonMark and other dialects allow backslash hard line breaks. See http://spec.commonmark.org/0.18/#example-527

This commit changes the semantics of data, but all tests pass.


neclepsio:
Sorry, wrong code.

PR #163, opened 2015-04-22
Implement backslash hard line break extension

neclepsio:
Many don't like double space because it's not visible. For this, CommonMark and other dialects allow backslash hard line breaks. See http://spec.commonmark.org/0.18/#example-527

I added an extension (EXTENSION_BACKSLASH_LINE_BREAK), added by default in Common profile.

All tests pass.


rtfb:
I would love to see some tests added to explicitly exercise this behavior.

neclepsio:
Done.

rtfb:
Awesome, thanks!

PR #174, opened 2015-06-05
HTML escaping flag

etw:


shurcooL:
@etw, can you please provide some motivation for why this flag would be useful to have in blackfriday? What is the intended use, or what problem is it meant to resolve?

etw:
@shurcooL  For security reasons, when markdown is provided by user.

shurcooL:
Are you already familiar with this section in the README?

https://github.com/russross/blackfriday#sanitize-untrusted-content

etw:
Yep, I am.But generating HTML and then parsing it and then sanitizing, it's well... unoptimized. It's much simplier and more effective to escape html code, when it's not needed.

shurcooL:
By unoptimized and efficient, I assume you're referring to the performance of the executed code rather than developer time spent developing, maintaining, and ensuring secure operation of libraries, right? Or how likely the generated output is to have security holes?

I agree it's potentially not as performant as parsing markdown and sanitizing it while rendering, but that approach is extremely hard to be sure is secure. It also couples unrelated concerns into one package. It was decided that separating the rendering from sanitization is the best way to ensure safety, and the performance cost is unlikely to become a bottleneck in most uses.

The main difficulty in trying to sanitize markdown while rendering it is that it's allowed to contain (potentially unsafe) HTML inside. If that weren't the case, it would be a lot more viable.

etw:
> The main difficulty in trying to sanitize markdown while rendering it is that it's allowed to contain (potentially unsafe) HTML inside.

It's up to service's decide, if html is allowed or not. If there is total ban for raw html tags in markdown input, there is nothing to sanitize there, except escaping all raw html code.
Using bluemonday (or any other separate sanitizer) is a valid argument, if we consider some html tags in markdown input as valid, but it's not the case.

rtfb:
> If there is total ban for raw html tags in markdown input, there is nothing to sanitize then, except escaping all raw html code.

You're absolutely right on that one, but the tricky part is to guarantee all the raw html code was really escaped. We tried it, and it's not as easy as it sounds.

etw:
Hmm... could you give an example, when escaping with attrEscape or https://golang.org/pkg/text/template/#HTMLEscapeString could be got around?

rtfb:
Sure, there you go:

```md
Paragraph<span>this one is escaped</span>\n\n<script>thisOneIsNot('CAN HAZ XSS PLZ?');</script>
```

But even if you go ahead and add a similar check to `BlockHtml` like you've added to `RawHtmlTag`, you still couldn't trust the produced html. This example gives some intuition about how one could produce a (valid, but) malicious markdown:

```md
![foo](http://ha.ckers.org/xss.js)
```

which would get rendered to:

```html
<p><img src=\"http://ha.ckers.org/xss.js\" alt=\"foo\" /></p>
```

etw:
Ok, got the point.

Issue #136, opened 2014-12-22
Comments wrongly converted in lists


duncombe:
I am using [Hugo](github.com/spf13/hugo), and encountered the bug detailed below. I am told blackfriday is the markdown parser used by Hugo, so I am reporting the bug here too.

Within a list (e.g., `<DL>` `</DL>`) in a markdown document, Hugo converts comments (`<!-- -->`) to `&lt;!&ndash; &ndash;&gt;`. Outside the list scope the comments are treated correctly and left as comments (`<!-- -->`). This is a bug.

I first observed this bug in definition lists, but it turns out all lists are affected, including markdown lists demarcated with `-` or `*`, not just HTML tagged lists, with the comments incorrectly rendered, turning them into displayed characters. The bug can be circumvented by separating the comments from the list using copious amounts of blank lines, but these blank lines then appear in the HTML affecting the layout of the list.

Refers  to spf13/hugo#729.

anthonyfok:
It is not just lists... Paragraphs too!  When I tested with https://github.com/russross/blackfriday-tool:

    Hello 

gives

> <p>Hello &lt;!&ndash; there &ndash;&gt;</p>

    <p>Hello &lt;!&ndash; there &ndash;&gt;</p>

This is somehow related to `smartDash()` and/or `smartDashLatex()` in `smartypants.go`, as well as `htmlComment()` in `block.go`.

yacoob:
Another case where it misbehaves, quoted from a [Hugo bug](https://github.com/spf13/hugo/issues/1080)

Source code like this:

```markdown
blahblah
<!--- foo -->
rhubarb
```

becomes

```html
blahblah
-&lt;!&ndash; foo &ndash;&gt;
rhubarb
```

If you separate comment in the source file by extra empty lines, the conversion is inhibited *but* extra <p> are added:

```markdown
blahblah
<!--- foo -->
rhubarb
```

becomes

```html
blahblah
</p>
<!-- foo -->
<p>
rhubarb
```

This is annoying :(

thatguystone:
This is probably the wrong way to fix it (it's a bit hacky), but it gets the job done: 9e5f8e534b83b9a15611f5af53600b9f2727aeae

bep:
Any plans to fix this?

rtfb:
@anthonyfok, am I right thinking that #190 is a workaround for this problem?

rtfb:
[...]

PR #190, opened 2015-08-04
Add HTML_SMARTYPANTS_DASHES for toggling smart dashes

anthonyfok:
There may be cases where an end user (like me) would like to use SmartyPants except the smart dashes, hence this pull request adds a new `HTML_SMARTYPANTS_DASHES` flag to allow the user to turn off just the smart dashes.  :-)

What do you think?  :-)


rtfb:
Missed this one. Looks good to me. Will merge if no objections are raised in couple days.

spf13:
Please merge :)

rtfb:
Well, I actually stumbled upon an (potential) objection myself... Since @anthonyfok submitted this one and he also was interested in #136, I was wondering whether this setting was meant to workaround the problem with #136. Or is this completely unrelated and only coincidentally involves two consecutive dashes? :-)

rtfb:
OK, I've read the motivation over at corresponding PR to Hugo and see no downsides. In it goes.

anthonyfok:
Wow, what a pleasant surprise!  Thank you so much @rtfb!
And thank you @spf13 for advocating for me!  :-)

(Sorry, I had been away, and hence the late reply to this thread.)

Cheers,
Anthony

Issue #201, opened 2015-10-06
Add the ability to extract a MarkDown-formatted ToC


TomOnTime:
MarkDown doesn't have a native "insert ToC here" directive.  Some systems have added one. Others, such as Gitlab, requires the ToC to be manually inserted.  I've been looking for a tool that would generate the ToC so that I can "sed" it into my MarkDown files.   Sadly I didn't find any, but I did find blackfriday which has 99% of the code required.

I hope you are open to pull requests that add new features.  I'll be sending you two (one for blackfriday, one for blackfriday-tool) which add a "-mdtoc" flag that simply outputs the ToC.  It isn't ideal (ideal would be that it inserts the ToC into the Markdown automatically).  However this is a good start and I'm sure many people will appreciate it.

Thanks!
Tom

qiansen1386:
Many markdown tools support following clause. Is it possible for us to add a support for that with an extension?
```
[TOC]
```

chaiko:
+1 for supporting `[TOC]`

fofen:
+1 for supporting `[TOC]`

PR #202, opened 2015-10-06
Add the ability to simply extract the ToC in Markdown format.

TomOnTime:
This is related to https://github.com/russross/blackfriday/issues/201

### DIFF
--- a/html.go
+++ b/html.go
@@ -21,6 +21,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"unicode"
 )
 
 // Html renderer configuration options.
@@ -34,6 +35,7 @@ const (
 	HTML_NOREFERRER_LINKS                      // only link with rel="noreferrer"
 	HTML_HREF_TARGET_BLANK                     // add a blank target
 	HTML_TOC                                   // generate a table of contents
+	HTML_TOC_MD                                // if generating table of contents, output MarkDown
 	HTML_OMIT_CONTENTS                         // skip the main contents (for a standalone table of contents)
 	HTML_COMPLETE_PAGE                         // generate a complete HTML page
 	HTML_USE_XHTML                             // generate XHTML output instead of HTML
@@ -482,9 +484,15 @@ func (options *Html) AutoLink(out *bytes.Buffer, link []byte, kind int) {
 }
 
 func (options *Html) CodeSpan(out *bytes.Buffer, text []byte) {
-	out.WriteString("<code>")
-	attrEscape(out, text)
-	out.WriteString("</code>")
+	if options.flags&HTML_TOC_MD == 0 {
+		out.WriteString("<code>")
+		attrEscape(out, text)
+		out.WriteString("</code>")
+	} else {
+		out.WriteString("`")
+		out.Write(text)
+		out.WriteString("`")
+	}
 }
 
 func (options *Html) DoubleEmphasis(out *bytes.Buffer, text []byte) {
@@ -728,9 +736,13 @@ func (options *Html) DocumentFooter(out *bytes.Buffer) {
 		}
 
 		// insert the table of contents
-		out.WriteString("<nav>\n")
-		out.Write(options.toc.Bytes())
-		out.WriteString("</nav>\n")
+		if options.flags&HTML_TOC_MD == 0 {
+			out.WriteString("<nav>\n")
+			out.Write(options.toc.Bytes())
+			out.WriteString("</nav>\n")
+		} else {
+			out.Write(options.toc.Bytes())
+		}
 
 		// corner case spacing issue
 		if options.flags&HTML_COMPLETE_PAGE == 0 && options.flags&HTML_OMIT_CONTENTS == 0 {
@@ -750,7 +762,38 @@ func (options *Html) DocumentFooter(out *bytes.Buffer) {
 
 }
 
+func makeAnchorText(text []byte) string {
+	// This is compatible with GitLab anchors.
+	var t []rune
+	s := strings.ToLower(string(text))
+	s = strings.Replace(s, " `-", "-", -1)
+	for _, r := range s {
+		switch {
+		case unicode.IsSpace(r) || r == '-':
+			t = append(t, '-')
+		case unicode.IsNumber(r) || unicode.IsLetter(r):
+			t = append(t, r)
+		}
+
+	}
+	return string(t)
+}
+
+func (options *Html) mdTocHeaderWithAnchor(text []byte, level int) {
+	options.toc.WriteString(strings.Repeat(" ", level-1))
+
+	options.toc.WriteString("- [")
+	options.toc.Write(text)
+	options.toc.WriteString("](#")
+	options.toc.WriteString(makeAnchorText(text))
+	options.toc.WriteString(")\n")
+}
+
 func (options *Html) TocHeaderWithAnchor(text []byte, level int, anchor string) {
+	if options.flags&HTML_TOC_MD != 0 {
+		options.mdTocHeaderWithAnchor(text, level)
+		return
+	}
 	for level > options.currentLevel {
 		switch {
 		case bytes.HasSuffix(options.toc.Bytes(), []byte("</li>\n")):
@@ -801,7 +844,7 @@ func (options *Html) TocFinalize() {
 		options.currentLevel--
 	}
 
-	if options.currentLevel > 0 {
+	if options.currentLevel > 0 && options.flags&HTML_TOC_MD == 0 {
 		options.toc.WriteString("</ul>\n")
 	}
 }
--- a/html_test.go
+++ b/html_test.go
@@ -1,0 +1,103 @@
+//
+// Blackfriday Markdown Processor
+// Available at http://github.com/russross/blackfriday
+//
+// Copyright © 2011 Russ Ross <russ@russross.com>.
+// Distributed under the Simplified BSD License.
+// See README.md for details.
+//
+
+//
+// Unit tests for html rendering
+//
+
+package blackfriday
+
+import (
+	"testing"
+)
+
+func doTestsMakeAnchor(t *testing.T, tests []string) {
+	// catch and report panics
+	var candidate string
+
+	for i := 0; i+1 < len(tests); i += 2 {
+		input := tests[i]
+		candidate = input
+		expected := tests[i+1]
+		actual := makeAnchorText([]byte(candidate))
+		if actual != expected {
+			t.Errorf("\nInput   [%#v]\nExpected[%#v]\nActual  [%#v]",
+				candidate, expected, actual)
+		}
+
+	}
+}
+
+func TestMakeAnchor(t *testing.T) {
+	var tests = []string{
+		"httpunit",
+		"httpunit",
+
+		"Architecture overview:",
+		"architecture-overview",
+
+		"http unit \"by hand\":",
+		"http-unit-by-hand",
+
+		"http unit \"by file\":",
+		"http-unit-by-file",
+
+		"TOML file format:",
+		"toml-file-format",
+
+		"Basic test parameters",
+		"basic-test-parameters",
+
+		"IP variables",
+		"ip-variables",
+
+		"Testing with regular expressions",
+		"testing-with-regular-expressions",
+
+		"Additional command line flags",
+		"additional-command-line-flags",
+
+		"`-filter string`",
+		"-filter-string",
+
+		"`-header string`",
+		"-header-string",
+
+		"`-ipmap string`",
+		"-ipmap-string",
+
+		"`-no10`",
+		"-no10",
+
+		"`-timeout duration`",
+		"-timeout-duration",
+
+		// This is being bug compatible with Gitlab.
+		// Some would say the result should be: "-v-and--vv"
+		"`-v` and `-vv`",
+		"-v-and-vv",
+
+		"Common Tasks",
+		"common-tasks",
+
+		"Add a rule for a new haproxy port.",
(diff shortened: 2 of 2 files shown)


rtfb:
Great, let's wait what @shurcooL has to say about it :-)

shurcooL:
I've looked over this and the proposed change to `sanitized_anchor_name`.

I don't think that changing `sanitized_anchor_name` is the best way to solve this task. The update is specific to how GitLab happens to perform their header title -> sanitized anchor name mapping. If someone else later wants to make their anchors compatible with another style associated with a different company, they'd have to change `sanitized_anchor_name` yet again and it'd break the new GitLabs-flavored behavior.

Since what's being added here is a new optional feature for specific users to opt into, I think the best way is to add a new field to [`Options`](https://godoc.org/github.com/russross/blackfriday#Options) type:

```Go
// Options represents configurable overrides and callbacks (in addition to the
// extension flag set) for configuring a Markdown parse.
type Options struct {
	...

	// SanitizedAnchorNameOverride is an optional func to override
	// the behavior for creating sanitized anchor names.
	SanitizedAnchorNameOverride SanitizedAnchorNameFunc
}

// SanitizedAnchorNameFunc creates sanitized anchor names.
type SanitizedAnchorNameFunc func(text string) string
```

By default (if `SanitizedAnchorNameOverride` is left as `nil`), the current `sanitized_anchor_name.Create` can be used to provide a simple default behavior.

If GitLab-specific behavior for creating anchor names is desired, that func can be set by the user to override and get the behavior they want.

What do you think of that approach?

The main reason I don't want to change `sanitized_anchor_name` is because the changes are not specific to a widely accepted standard but rather an arbitrary set of choices that are specific to GitLab only, as far as I can tell. Given that, I expect there would be similar variations elsewhere. `sanitized_anchor_name` is mean to be a very simple default. But if you disagree here, please let me know.

TomOnTime:
Sounds good!  Though I haven't tested GitHub or other systems to see if they use the same algorithm.

TomOnTime:
Ok, I did some tests with Github and it does have differences:

https://github.com/TomOnTime/tomontime/wiki/test_headers
https://gitlab.com/tal/gitlab-markdown-test/wikis/gitlab_headers

Here are the cases where things are generated differently:
https://gist.github.com/TomOnTime/eff1439aa0682f34653b

How about I add CreateGitLab() and CreateGitHub()  to sanitized_anchor_name/main.go and make main_test.go have both test suites?

TomOnTime:
Assuming the changes to https://github.com/shurcooL/sanitized_anchor_name are accepted, this should now let you select the anchor sanitizer of your choosing.

markdown.go
```diff
@@ -374,6 +387,12 @@ func MarkdownOptions(input []byte, renderer Renderer, opts Options) []byte {
 		p.notes = make([]*reference, 0)
 	}
 
+	if opts.SanitizedAnchorNameOverride == nil {
+		p.sanitizedAnchorNameOverride = sanitized_anchor_name.Create
```
shurcooL: This logic could be done in one place (in html.go) instead of two.

html.go
```diff
@@ -106,12 +114,12 @@ const (
...
 // title and css are only used when HTML_COMPLETE_PAGE is selected.
-func HtmlRenderer(flags int, title string, css string) Renderer {
-	return HtmlRendererWithParameters(flags, title, css, HtmlRendererParameters{})
+func HtmlRenderer(flags int, title string, css string, sanitize SanitizedAnchorNameFunc) Renderer {
+	return HtmlRendererWithParameters(flags, title, css, sanitize, HtmlRendererParameters{})
 }
 
 func HtmlRendererWithParameters(flags int, title string,
-	css string, renderParameters HtmlRendererParameters) Renderer {
+	css string, sanitize SanitizedAnchorNameFunc, renderParameters HtmlRendererParameters) Renderer {
```
shurcooL: Why did you choose to change the signature of `HtmlRendererWithParameters` instead of adding the `SanitizedAnchorNameFunc` as a new field to `HtmlRendererParameters`?

TomOnTime:
> Why did you choose to change the signature of HtmlRendererWithParameters instead of adding the
> SanitizedAnchorNameFunc as a new field to HtmlRendererParameters?

Short answer: I wanted to but I didn't see how to do it and still get the info from blackfriday-tool to that place.  Hints?

rtfb:
BTW, it suddenly struck me that this functionality fits ideally in https://github.com/shurcooL/markdownfmt, while in Blackfriday it introduces a completely new form of output. Also, you wouldn't even need a new flag in markdownfmt, only make it respect `HTML_OMIT_CONTENTS`.

TomOnTime:
@rtfb 

I've push a change that addresses the concerns (HtmlRendererWithParameters no longer has a new signature, etc).  Sadly it breaks one test in a way that I can't figure out how to fix.  Do you have any suggestions?

As far as adding the functionality to https://github.com/shurcooL/markdownfmt instead, I'd be interested in @shurcooL's opinion.  It seems like that would require a lot of code changes.

@shurcooL I see two ways to make this work as part of markdownfmt instead.  One would be to simply have HTML_OMIT_CONTENTS omit most output, but when outputting Headers instead output the ToC entry.  The other would be to look for a token pair such as [TOC][/TOC] and replace the contents with the ToC.  The latter seems like a better way to do it, but a lot more work (and beyond my understanding of how to add that function).  Thoughts?

shurcooL:
About https://github.com/shurcooL/markdownfmt, the `markdownfmt` command is designed not to have any formatting options (just like `gofmt` doesn't have formatting options), so I don't think it would work. It's also outside the scope of that project. You're welcome to fork it and modify it to be a dedicated tool for what you're doing, if you want.

---

> Assuming the changes to https://github.com/shurcooL/sanitized_anchor_name are accepted, this should now let you select the anchor sanitizer of your choosing.

I've reviewed the latest changes at https://github.com/shurcooL/sanitized_anchor_name/pull/2#issuecomment-149087410, please see that comment.

As a followup, unless I'm mistaken, no changes to the current `sanitized_anchor_name` package need to be accepted for this PR to be able to go through. Additional sanitize-anchor-name functionality can be in any Go package, not necessarily that one. Is that correct? In https://github.com/shurcooL/sanitized_anchor_name/pull/2#issuecomment-149087410, I've outlined the reasons why I don't think it's the best home for it.

rtfb:
So what's the status of this, @TomOnTime? Are you satisfied with how things ended up partitioned? This PR looks good now, I can merge if you think it's what it should be.

block_test.go
```diff
@@ -35,6 +37,10 @@ func runnerWithRendererParameters(parameters HtmlRendererParameters) func(string
 		htmlFlags := 0
 		htmlFlags |= HTML_USE_XHTML
 
+		if parameters.SanitizedAnchorNameOverride == nil {
+			parameters.SanitizedAnchorNameOverride = sanitized_anchor_name.Create
+		}
+
 		renderer := HtmlRendererWithParameters(htmlFlags, "", "", parameters)
```
shurcooL: Are the 4 added lines above needed? `HtmlRendererWithParameters` already sets the default value for `SanitizedAnchorNameOverride` if it's `nil`.

TomOnTime:
I haven't had a chance to look at this.  I'll get to it in the next few days.

TomOnTime:
I made the changes as requested. Also, the GitLab and GitHub functionality was split into markdownutils, so I've adjusted the code as required.  This is ready for merge.

html.go
```diff
@@ -70,6 +74,9 @@ type HtmlRendererParameters struct {
 	HeaderIDPrefix string
 	// If set, add this text to the back of each Header ID, to ensure uniqueness.
 	HeaderIDSuffix string
+	// SanitizedAnchorNameOverride is an optional func to override
+	// the behavior for creating sanitized anchor names.
+	SanitizedAnchorNameOverride SanitizedAnchorNameFunc
```
shurcooL: This PR should be about adding the _option_ to change the default. These 3 lines are good.

html.go
```diff
@@ -34,6 +37,7 @@ const (
 	HTML_NOREFERRER_LINKS                      // only link with rel="noreferrer"
 	HTML_HREF_TARGET_BLANK                     // add a blank target
 	HTML_TOC                                   // generate a table of contents
+	HTML_TOC_MD                                // if generating table of contents, output MarkDown
```
shurcooL: This library spells Markdown as "Markdown", not "MarkDown". This should be kept consistent.

It's not very clear what this flag does. Output Markdown instead of what?

html.go
```diff
@@ -123,6 +130,10 @@ func HtmlRendererWithParameters(flags int, title string,
 		renderParameters.FootnoteReturnLinkContents = `<sup>[return]</sup>`
 	}
 
+	if renderParameters.SanitizedAnchorNameOverride == nil {
+		renderParameters.SanitizedAnchorNameOverride = markdownutils.CreateGitLabAnchor
```
shurcooL: What's the motivation to change the default? This seems unnecessary. Users will be able to set it to `markdownutils.CreateGitLabAnchor` or another custom func by using the `SanitizedAnchorNameOverride` parameter.

shurcooL:
> * Use github.com/TomOnTime/markdownutils instead of sanitized_anchor_name

I don't think this PR should be making that change.

html.go
TomOnTime: Fixed.

html.go
TomOnTime: Removed.  My intention was to avoid a nil reference, but testing the code with those 3 lines removed works so I guess it wasn't needed.

shurcooL: Perhaps you misunderstood. I meant that `blackfriday` should continue to use `sanitized_anchor_name.Create` as the default anchor sanitizer. This PR should not change the default to `markdownutils.CreateGitLabAnchor`, that should be done in your own code that imports and uses `blackfriday`.

TomOnTime:
Ah!  Now I get it.  The dependency isn't needed, so it can removed.  (Now I understand what you were saying before)

block.go
```diff
@@ -243,7 +243,11 @@ func (p *parser) prefixHeader(out *bytes.Buffer, data []byte) int {
 	}
 	if end > i {
 		if id == "" && p.flags&EXTENSION_AUTO_HEADER_IDS != 0 {
-			id = sanitized_anchor_name.Create(string(data[i:end]))
+			if p.sanitizedAnchorNameOverride == nil {
+				id = sanitized_anchor_name.Create(string(data[i:end]))
+			} else {
+				id = p.sanitizedAnchorNameOverride(string(data[i:end]))
+			}
```
shurcooL: I think it'd be cleaner to ensure `p.sanitizedAnchorNameOverride` is never nil (by giving it a default value when there's no override), so the code here can be just:

```Go
id = p.sanitizedAnchorNameOverride(string(data[i:end]))
```

At that point, you'd want to rename `sanitizedAnchorNameOverride` to just `sanitizedAnchorName`.

If an override is provided, its value will be set to:

```Go
sanitizedAnchorName = sanitized_anchor_name.Create
```

block.go
```diff
@@ -1346,7 +1350,11 @@ func (p *parser) paragraph(out *bytes.Buffer, data []byte) int {
 
 				id := ""
 				if p.flags&EXTENSION_AUTO_HEADER_IDS != 0 {
-					id = sanitized_anchor_name.Create(string(data[prev:eol]))
+					if p.sanitizedAnchorNameOverride == nil {
+						id = sanitized_anchor_name.Create(string(data[prev:eol]))
+					} else {
+						id = p.sanitizedAnchorNameOverride(string(data[prev:eol]))
+					}
```
shurcooL: Same here, this could become just:

```Go
id = p.sanitizedAnchorName(string(data[i:end]))
```

html.go
```diff
@@ -483,9 +488,15 @@ func (options *Html) AutoLink(out *bytes.Buffer, link []byte, kind int) {
...
-	out.WriteString("</code>")
+	if options.flags&HTML_TOC_MD == 0 {
+		out.WriteString("<code>")
+		attrEscape(out, text)
+		out.WriteString("</code>")
+	} else {
+		out.WriteString("`")
+		out.Write(text)
+		out.WriteString("`")
+	}
```
shurcooL: I don't like the extra complexity this brings, but I don't see a different way of doing this if we're to support this feature.

markdown.go
```diff
@@ -380,6 +392,8 @@ func MarkdownOptions(input []byte, renderer Renderer, opts Options) []byte {
 		p.notes = make([]*reference, 0)
 	}
 
+	p.sanitizedAnchorNameOverride = opts.SanitizedAnchorNameOverride
```
shurcooL: This would be where you'd check if `opts.SanitizedAnchorNameOverride` is `nil` and use an internal default value if so:

```Go
if opts.SanitizedAnchorNameOverride != nil {
    p.sanitizedAnchorName = opts.SanitizedAnchorNameOverride
} else {
    p.sanitizedAnchorName = sanitized_anchor_name.Create
}
```

shurcooL:
> Ah! Now I get it. The dependency isn't needed, so it can removed. (Now I understand what you were saying before)

Great! The latest version is starting to look a lot better! I've left a few more suggestions, what do you think about them?

TomOnTime:
I'll try to get to it this week.

jonny-improbable:
Did anything ever come of this feature?

TomOnTime:
Sorry, I'm going to have to decline to continue working on this.  If someone else is interested, I'd be glad to help.  Sadly the project I was going to use this for is no longer happening and I won't have time to finish the PR.

shurcooL:
Thanks for letting us know @TomOnTime, and thank you for the work to get it to this point!

PR #205, opened 2015-10-12
Fix bug parsing emphasis

### LINKED ISSUE #156: **`code`** doesn't render the same in all situations
Here is my input:
```
**`/usr`** : this folder contains other important folders such as `/usr/bin`, `/usr/lib` and `/usr/share`.

**`/usr`** : 

this folder contains other important folders such as `/usr/bin`, `/usr/lib` and `/usr/share`.
```
And here is the html output:
```
<p>**<code>/usr</code>**: this folder contains other important folders such as <code>/usr/bin</code>, <code>/usr/lib</code> and <code>/usr/share</code>.</p>

<p><strong><code>/usr</code></strong>:</p>

<p>this folder contains other important folders such as <code>/usr/bin</code>, <code>/usr/lib</code> and <code>/usr/share</code>.</p>
```
Note the ** in the first case stay **, while in the second case the turn into `<strong>`.

Note that I've encountered this while using [hugo](https://github.com/spf13/hugo), version 0.13.

rtfb:
Start searching for emphasis character at 0th index instead of 1st.
Fixes a corner case with doubly emphasised code span followed by
another code span on the same line.

Changes interpretation of improperly nested emphasis, hence the change
in TestEmphasisMix().

Closes #156.

### DIFF
--- a/inline.go
+++ b/inline.go
@@ -910,7 +910,7 @@ func isMailtoAutoLink(data []byte) int {
 
 // look for the next emph char, skipping other constructs
 func helperFindEmphChar(data []byte, c byte) int {
-	i := 1
+	i := 0
 
 	for i < len(data) {
 		for i < len(data) && data[i] != c && data[i] != '`' && data[i] != '[' {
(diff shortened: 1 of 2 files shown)


shurcooL:
LGTM.

PR #206, opened 2015-10-12
Fix footnote followed by a reference style link

### LINKED ISSUE #164: Footnote followed by links doesn’t work
This is an issue similar to #158, but with footnotes and links.
With:

```
Bla bla [^1] [Access-Control-Allow-Origin][allow]

---

[^1]: This is a footnote

[allow]: http://www.w3.org/TR/2008/WD-access-control-20080912/#access-control-allow-origin
```

It only translates the link (I get the same result if I use an inline link instead), and preserves the footnote like plain text:

> Bla bla [^1]&#x2063; [Access-Control-Allow-Origin](http://www.w3.org/TR/2008/WD-access-control-20080912/#access-control-allow-origin)

I just needed to use an invisible separator here (`&#x2063;`), between the footnote and link, for get it correctly processed.

rtfb:
When parsing a deferred footnote, we already know it's a footnote from
the '[^' part, so we can use that to hit a proper switch branch
(default) a bit later on.

Closes #164.

### DIFF
--- a/inline.go
+++ b/inline.go
@@ -191,6 +191,13 @@ const (
 	linkInlineFootnote
 )
 
+func isReferenceStyleLink(data []byte, pos int, t linkType) bool {
+	if t == linkDeferredFootnote {
+		return false
+	}
+	return pos < len(data)-1 && data[pos] == '[' && data[pos+1] != '^'
+}
+
 // '[': parse a link or an image or a footnote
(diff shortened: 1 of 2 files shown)


shurcooL:
LGTM, but I am not very familiar with reference style links (I've never used them).

PR #208, opened 2015-10-13
Fix escaping asterisks within emphasis

### LINKED ISSUE #18: Escaping astericks does not work
Typing 
     \*What is A\* all about?*
 should be rendered in all italics, but isn't
Github also messes it up, cool. GFM != Markdown

rtfb:
First check for escaped character, *then* look if i-th character is an
emphasis character.

Closes #18.


shurcooL:
LGTM.

PR #210, opened 2015-10-23
support for embeding css into the generated-html

oec:
Hi,

thanks a ton for blackfriday!  I had the need to embed a css-file into the generated HTML-output (so I have a fully self-contained, single file to use as attachment, f.e.), so I added an option to the flags to communicate that to the renderer and read the css-file and write it inside <style>-tags.

I hope you find that useful.  There will be a corresponding pull-request for the blackfriday-tool.

Cheers,
  Özgür

### DIFF
--- a/html.go
+++ b/html.go
@@ -18,6 +18,8 @@ package blackfriday
 import (
 	"bytes"
 	"fmt"
+	"io/ioutil"
+	"log"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


rtfb:
I feel undecided about this one. There's nothing wrong about it, so we could take it. But its niche looks so narrow and it's easy to work around outside of blackfriday, so it feels like taking the code that won't be maintained.

I think to decline this one, unless somebody wants to advocate for it.

Issue #122, opened 2014-10-19
Multiple fenced code blocks inside a ">"-prefixed blockquote rendered incorrectly


anthonyfok:
Hello Russ,

Thank you very much for writing Blackfriday, as it is a very important Markdown engine for Go that makes easy blogging and super-fast rendering with Go applications like [Hugo](http://gohugo.io/) a possibility.

I think I may have stumbled across a bug in Blackfriday when I tried to nest multiple fenced (```) code blocks inside a ">"-prefixed blockquote section when I was experimenting with writing a blog entry a few days ago.

Here is a modified/shortened example:

#### Markdown source code:

----

<pre>
> Normally, `ls -l` shows the file date/time rounded to the nearest minute.  For example:
>
> ```sh
user@hostname:/etc/cups/ppd$ ls -l
-rw-r--r-- 1 root root 21279 Oct  8 18:07 FUJI-XEROX-DocuPrint-CM215-fw.ppd
-rw-r----- 1 root lp   21519 Oct  8 18:07 Xerox_WorkCentre_6015NI.ppd
```
>
> To find out the precise time when these PPD files were updated, we may use add the `--full-time` option to `ls`:
>
> ```sh
user@hostname:/etc/cups/ppd$ ls -l --full-time
-rw-r--r-- 1 root root 21279 2014-10-08 18:07:38.912867452 -0600 FUJI-XEROX-DocuPrint-CM215-fw.ppd
-rw-r----- 1 root lp   21519 2014-10-08 18:07:39.032869071 -0600 Xerox_WorkCentre_6015NI.ppd
```
>
> This may be used for comparing with the timestamps in /var/log/dpkg.log to pinpoint the exact package which changed the file permission and ownership of Xerox_WorkCentre_6015NI.ppd of my default printer.
</pre>

----

#### Expected result (rendered correctly by GitHub's Markdown engine):

----

> Normally, `ls -l` shows the file date/time rounded to the nearest minute.  For example:
>
> ```sh
user@hostname:/etc/cups/ppd$ ls -l
-rw-r--r-- 1 root root 21279 Oct  8 18:07 FUJI-XEROX-DocuPrint-CM215-fw.ppd
-rw-r----- 1 root lp   21519 Oct  8 18:07 Xerox_WorkCentre_6015NI.ppd
```
>
> To find out the precise time when these PPD files were updated, we may use add the `--full-time` option to `ls`:
>
> ```sh
[...]

Issue #45, opened 2013-12-17
Fenced Code Blocks without a blank line before.


dmitshur:
GitHub Flavored Markdown has a section on [fenced code blocks](https://help.github.com/articles/github-flavored-markdown#fenced-code-blocks), where it says "Keep in mind that both types of code blocks need to have a blank line before them".

github.com, however, ignores that statement, and renders markdown without such a blank line correctly (i.e. as if the blank line were present). `blackfriday` currently does not follow that behavior (tested with `blackfriday.MarkdownCommon`, and renders such markdown in an odd manner. Should it be changed to match that of GitHub?

The follow markdown reproduces the issue:

``````
some text without a blank line afterwards
```Go
someCode()
```
``````

For a larger example, see this Markdown source [1], how github.com displays it [2], and how `blackfriday.MarkdownCommon` renders it [3].

[1] - https://raw.github.com/shurcooL/go-goon/8ddcefebec68d2dbcbac5225bf8760fbd4598c47/README.md
[2] - https://github.com/shurcooL/go-goon/blob/8ddcefebec68d2dbcbac5225bf8760fbd4598c47/README.md
[3] - http://dl.dropboxusercontent.com/u/8554242/available-for-2-weeks/fenced_code_blocks_blackfriday.html

PR #211, opened 2015-10-29
Fix issue 122: fenced code blocks inside blockquotes

rtfb:
Improves fenced code block detection in paragraphs and blockquotes. Directly addresses #122, but refixes #45 along the way. More details in respective commit messages.

### DIFF
--- a/block_test.go
+++ b/block_test.go
@@ -1065,6 +1065,78 @@ func TestFencedCodeBlock(t *testing.T) {
 	doTestsBlock(t, tests, EXTENSION_FENCED_CODE)
 }
 
+func TestFencedCodeInsideBlockquotes(t *testing.T) {
+	var tests = []string{
+		"> ```go\n" +
+			"package moo\n\n" +
+			"```\n",
+		`<blockquote>
+<pre><code class="language-go">package moo
+
+</code></pre>
+</blockquote>
+`,
+		// -------------------------------------------
+		"> foo\n" +
+			"> \n" +
+			"> ```go\n" +
+			"package moo\n" +
+			"```\n" +
+			"> \n" +
+			"> goo.\n",
+		`<blockquote>
+<p>foo</p>
+
+<pre><code class="language-go">package moo
+</code></pre>
+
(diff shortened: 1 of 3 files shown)


markdown.go
```diff
@@ -393,7 +393,6 @@ func firstPass(p *parser, input []byte) []byte {
 		tabSize = TAB_SIZE_EIGHT
```
shurcooL: 5 lines above, the comment for `firstPass` func still says:

```Go
// - add missing newlines before fenced code blocks
```

That should be removed.

block_test.go
```diff
@@ -1065,6 +1065,78 @@ func TestFencedCodeBlock(t *testing.T) {
...
+<p>goo.</p>
+
+<pre><code class="language-go">package zoo
+</code></pre>
+
+<p>woo.</p>
+</blockquote>
+`,
+	}
+	doTestsBlock(t, tests, EXTENSION_FENCED_CODE)
```
shurcooL: I find these tests really hard to read.

I will propose an alternative way to represent those things in a PR (into this PR), which I find slightly more readable. I hope you will find it more readable too, then please take it.

I also think it's important to cover both forms of quoted fenced blocks and make sure they correctly preserve leading space. I think the current tests don't cover the 2nd form (but I could be mistaken, because it's really hard to read the test above):

	> plain quoted text
	> ```fenced
	code
	 with leading single space correctly preserved
	okay
	```
	> rest of quoted text

---

	> plain quoted text
	> ```fenced
	> code
	>  with leading single space correctly preserved
	> okay
	> ```
	> rest of quoted text

Both those inputs should generate (and they correctly do):

![image](https://cloud.githubusercontent.com/assets/1924134/10865861/b39e2edc-7fd7-11e5-93ab-bb5e8214547e.png)

shurcooL: See #214.

shurcooL:
I've left some comments. Everything else looks good.

markdown.go
rtfb: Good catch, thanks!

PR #214, opened 2015-10-31
Make test values more readable, add additional test cases.

shurcooL:
This is a PR into branch `issue-122`, not `master`.

Addresses https://github.com/russross/blackfriday/pull/211/files#r43576434.

### DIFF
--- a/block_test.go
+++ b/block_test.go
@@ -14,6 +14,7 @@
 package blackfriday
 
 import (
+	"strings"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


shurcooL:
Hmm, I've realized the first and remaining lines will only be aligned if you're using tab width of 4.

E.g., https://github.com/russross/blackfriday/pull/214/files?ts=4 is aligned, but https://github.com/russross/blackfriday/pull/214/files?ts=8 is not.

rtfb:
Thanks! Even with tab width 8 the test lines get a bit more readable by dropping the `\n`, so it's still an improvement.

shurcooL:
Great. :)

PR #215, opened 2015-11-02
Fix nested footnotes

### LINKED ISSUE #193: Nested footnotes are processed, but not rendered
When nesting footnotes (e.g. adding a footnote to the text of another footnote), the reference link is created, but the footnote `<li>` isn't added to the rendered HTML.

Example:
```
This is standard text.[^fn1]

[^fn1]:
  This is a footnote. It has a footnote of its own.[^fn2]

[^fn2]:
  This is a second footnote.
```

Output:
```html
<p>This is standard text.<sup class="footnote-ref" id="fnref:b128b7694e4de962e0088bcb1bcb254f:fn1"><a rel="footnote" href="#fn:b128b7694e4de962e0088bcb1bcb254f:fn1">1</a></sup></p>
<div class="footnotes">

<hr>

<ol>
<li id="fn:b128b7694e4de962e0088bcb1bcb254f:fn1">This is a footnote. It has a footnote of its own.<sup class="footnote-ref" id="fnref:b128b7694e4de962e0088bcb1bcb254f:fn2"><a rel="footnote" href="#fn:b128b7694e4de962e0088bcb1bcb254f:fn2">2</a></sup>
 <a class="footnote-return" href="#fnref:b128b7694e4de962e0088bcb1bcb254f:fn1">↩</a></li>
</ol>
</div>
```

rtfb:
This is both nasty and neat at the same time. All the code could handle nested footnotes just fine, the only place that was not working was the final loop that printed the list. The loop was in a range form, which couldn't account for another footnote being inserted while processing existing ones. Changing the loop to the iterative form solves that.

Fixes #193.

### DIFF
--- a/markdown.go
+++ b/markdown.go
@@ -452,8 +452,9 @@ func secondPass(p *parser, input []byte) []byte {
 	if p.flags&EXTENSION_FOOTNOTES != 0 && len(p.notes) > 0 {
 		p.r.Footnotes(&output, func() bool {
 			flags := LIST_ITEM_BEGINNING_OF_LIST
-			for _, ref := range p.notes {
+			for i := 0; i < len(p.notes); i += 1 {
 				var buf bytes.Buffer
+				ref := p.notes[i]
 				if ref.hasBlock {
 					flags |= LIST_ITEM_CONTAINS_BLOCK
 					p.block(&buf, ref.title)
--- a/markdown.go
+++ b/markdown.go
@@ -20,6 +20,7 @@ package blackfriday
 
 import (
 	"bytes"
+	"fmt"
 	"strings"
 	"unicode/utf8"
 )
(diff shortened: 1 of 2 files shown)


markdown.go
```diff
@@ -452,8 +453,9 @@ func secondPass(p *parser, input []byte) []byte {
 	if p.flags&EXTENSION_FOOTNOTES != 0 && len(p.notes) > 0 {
 		p.r.Footnotes(&output, func() bool {
 			flags := LIST_ITEM_BEGINNING_OF_LIST
-			for _, ref := range p.notes {
+			for i := 0; i < len(p.notes); i += 1 {
 				var buf bytes.Buffer
+				ref := p.notes[i]
```
shurcooL: Style suggestion, I would move `ref := p.notes[i]` to the the very first thing after `for i := 0; i < len(p.notes); i += 1 {`, since you're effectively replacing the `ref` from `for _, ref := range p.notes {`. I think it's slightly cleaner that way, but I don't feel very strongly about it if you disagree.

shurcooL:
I think it's neat. Using iterative form and strictly adding to the end of the current list looks like correct behavior to me.

shurcooL:
LGTM.

PR #216, opened 2015-11-03
Fix footnote following an exclamation point

### LINKED ISSUE #194: Footnotes are not rendered following an exclamation point
When placing a footnote after a sentence with an exclamation point, the footnote is not rendered properly.

Example:
```
Something exciting![^fn1]

[^fn1]:
  This is a footnote.
```

Output:
```html
<p>Something exciting![^fn1]</p>
```

rtfb:
Link parser interpreted the sequence "![^foo]" as an image, but if
footnote extension is enabled, it's quite clear that it should be
interpreted as a footnote following something with an exclamation point
at the end.

Closes #194.

### DIFF
--- a/inline.go
+++ b/inline.go
@@ -205,24 +205,26 @@ func link(p *parser, out *bytes.Buffer, data []byte, offset int) int {
 		return 0
 	}
 
-	// [text] == regular link
+	var t linkType
+	switch {
+	// special case: ![^text] == deferred footnote (that follows something with
+	// an exclamation point)
+	case p.flags&EXTENSION_FOOTNOTES != 0 && len(data)-1 > offset && data[offset+1] == '^':
+		t = linkDeferredFootnote
 	// ![alt] == image
+	case offset > 0 && data[offset-1] == '!':
+		t = linkImg
 	// ^[text] == inline footnote
 	// [^refId] == deferred footnote
-	var t linkType
-	if offset > 0 && data[offset-1] == '!' {
-		t = linkImg
-		// if footnotes extension is ON and we've seen "![^", then it's not an
-		// image, it's a deferred footnote:
-		if p.flags&EXTENSION_FOOTNOTES != 0 && len(data)-1 > offset && data[offset+1] == '^' {
-			t = linkDeferredFootnote
-		}
-	} else if p.flags&EXTENSION_FOOTNOTES != 0 {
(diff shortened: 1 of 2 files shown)


inline.go
```diff
@@ -212,6 +212,11 @@ func link(p *parser, out *bytes.Buffer, data []byte, offset int) int {
 	var t linkType
 	if offset > 0 && data[offset-1] == '!' {
 		t = linkImg
+		// if footnotes extension is ON and we've seen "![^", then it's not an
+		// image, it's a deferred footnote:
+		if p.flags&EXTENSION_FOOTNOTES != 0 && len(data)-1 > offset && data[offset+1] == '^' {
+			t = linkDeferredFootnote
+		}
```
shurcooL: There is a comment right above:

```Go
// [text] == regular link
// ![alt] == image
// ^[text] == inline footnote
// [^refId] == deferred footnote
```

I wonder if it's worth adding this special case to that list, i.e.:

```Go
// ![^refId] == deferred footnote
```

shurcooL: Behavior-wise this looks good to me. However, the current indentation style of the code makes this feel more like a hack/workaround than a long term solution (which it kinda is, since we're first setting `t` to `linkImg`, then over-writing it to be a `linkDeferredFootnote`, and also relying on a simple heuristic that most image URLs do not start with `^` as the first character).

Consider rewriting this if statement, which currently has 2 conditions `if offset > 0 && data[offset-1] == '!'` and `else if p.flags&EXTENSION_FOOTNOTES != 0`, into a switch statement with 3 conditions:

```Go
switch {
// ![^refId] == deferred footnote
case p.flags&EXTENSION_FOOTNOTES != 0 && len(data)-1 > offset && data[offset+1] == '^':
	// If footnotes extension is ON and we've seen "![^", then it's not an
	// image, it's a deferred footnote.
	t = linkDeferredFootnote

// ![alt] == image
case offset > 0 && data[offset-1] == '!':
	t = linkImg

// ^[text] == inline footnote
// [^refId] == deferred footnote
case p.flags&EXTENSION_FOOTNOTES != 0:
	if offset > 0 && data[offset-1] == '^' {
		t = linkInlineFootnote
	} else if len(data)-1 > offset && data[offset+1] == '^' {
		t = linkDeferredFootnote
	}
}
```

I think that would be more readable and maintainable. What do you think?

rtfb:
Good idea. Thanks!

PR #217, opened 2015-11-10
Use more idiomatic form for set of strings.

shurcooL:
This is a better style for a set, since each value can only be present or absent.

With bool as value type, each value may be absent, or true or false. It also uses slightly more memory.

Originated from https://github.com/russross/blackfriday/commit/f1ce82cb01f4e31e5c840eb8ef3cf2c0e7ab978f#commitcomment-14283676.


rtfb:
LGTM

PR #221, opened 2015-11-23
more better URL fragments

MiCHiLU:
see:
http://stackoverflow.com/questions/11155503/why-does-wikipedia-use-a-modified-percent-encoding-in-their-url-fragments

### DIFF
--- a/block.go
+++ b/block.go
@@ -15,6 +15,8 @@ package blackfriday
 
 import (
 	"bytes"
+	"net/url"
+	"strings"
 
 	"github.com/shurcooL/sanitized_anchor_name"
 )
@@ -243,7 +245,7 @@ func (p *parser) prefixHeader(out *bytes.Buffer, data []byte) int {
 	}
 	if end > i {
 		if id == "" && p.flags&EXTENSION_AUTO_HEADER_IDS != 0 {
-			id = sanitized_anchor_name.Create(string(data[i:end]))
+			id = strings.Replace(url.QueryEscape(sanitized_anchor_name.Create(string(data[i:end]))), "%", ".", -1)
 		}
 		work := func() bool {
 			p.inline(out, data[i:end])


shurcooL:
Can you elaborate on what you're trying to change? Give an example of what a fragment looked like previously, and after your proposed change.

I am guessing it has to do with how `sanitized_anchor_name.Create` operates on unicode characters. Its output is only letters, numbers and dash characters.

If there's an issue/enhancement to be made, there's a chance it might be better to have it inside `sanitized_anchor_name` rather than push the responsibility onto the callers. But I need to understand the issue better before I can see if that's the case.

MiCHiLU:
The `sanitized_anchor_name.Create` is do not nothing for the multibyte characters.
It will skip process and returns a multibyte character when larger than char code of Latin.
https://github.com/shurcooL/sanitized_anchor_name/blob/master/main.go#L18

It needs the URL based escape/quote for the URI that includes the multibyte characters.
https://en.wikipedia.org/wiki/Percent-encoding

shurcooL:
> The `sanitized_anchor_name.Create` is do not nothing for the multibyte characters.
> It will skip process and returns a multibyte character when larger than char code of Latin.

That's right, it assumes unicode input, and outputs unicode letters and numbers mostly unmodified, with the exception of turning letters to their lower case equivalent.

There's a test case for unicode [here](https://github.com/shurcooL/sanitized_anchor_name/blob/10ef21a441db47d8b13ebcc5fd2310f636973c77/main_test.go#L34).

> It needs the URL based escape/quote for the URI that includes the multibyte characters.

I will look for a spec or other source to confirm this.

If you could provide a browser or test case where you see it failing, that'd helpful.

shurcooL:
I wanted to look at how html/template escapes links with unicode (in the path, the query, and the fragment). Here's the result:

http://play.golang.org/p/yyqGBU93VN

```
t := template.Must(template.New("").Parse(`Here's a <a href="{{.}}">link</a>.`))
err := t.Execute(os.Stdout, template.URL("http://www.example.com/世界?q=世界#世界"))
if err != nil {
	log.Fatalln(err)
}
```

Output:

```HTML
Here's a <a href="http://www.example.com/%e4%b8%96%e7%95%8c?q=%e4%b8%96%e7%95%8c#%e4%b8%96%e7%95%8c">link</a>.
```

Here's how that renders in Chrome:

http://virtivia.com:27080/1agb5wu1btpas.html

image

shurcooL:
FWIW, here's a version with the fragment path unsanitized. It seems to work too:

```
Here's a <a href="http://www.example.com/%e4%b8%96%e7%95%8c?q=%e4%b8%96%e7%95%8c#世界">link</a>.
```

http://virtivia.com:27080/osxjmct6z5ef.html

image

rtfb:
If I'm reading this right, replacing percents with dots is a wrong thing to do in a general purpose code.

Calling `url.QueryEscape` seems like a good idea though, but it has to be called in both places where `sanitized_anchor_name.Create` is called. @shurcooL, do you think we should do percent-ecoding inside sanitized_anchor_name or outside?

shurcooL:
If it should be done anywhere, I think it should be done inside `sanitized_anchor_name.Create`.

However, I'm not sure `url.QueryEscape` should be used exactly. It's for escaping strings so they're safe to place inside a URL query. But `sanitized_anchor_name.Create` is for escaping things so they're safe to place inside URL fragments.

The best way I'd like to approach this is if someone can find a specification that documents what characters are allowed/not allowed in URL fragments and how they should be escaped, then we can can change it. I'd rather not make changes without a spec.

rtfb:
The way I'm reading [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt), fragments should be escaped the same way as other bits.

The section about percent encoding refers to a "component":

> 2.1.  Percent-Encoding
>
>   A percent-encoding mechanism is used to represent a data octet in a
>   component when that octet's corresponding character is outside the
>   allowed set [...]

and then syntax section clearly defines fragment to be one the components.

> \3.  Syntax Components
>
>   The generic URI syntax consists of a hierarchical sequence of
>   components referred to as the scheme, authority, path, query, and
>   fragment.

Combining these bits makes me think there should be no difference in percent escaping various pieces of URIs.

shurcooL:
That's very helpful, thanks!

I'll read it more closely soon and post an update here.

shurcooL:
I've been trying to get some time to look at this, but unfortunately haven't had an opportunity so far. I will hopefully get to it soon.

shurcooL:
Apologies for it taking me so long to get back. I've finally spent the time needed to properly look at the issue and can post an update now. The issue turned out more complex than I expected.

Here's what I found.

There is indeed a small issue in current code of blackfriday. It is very minor and not noticeable. This PR serves as a report of the issue (in addition to proposing a solution).

I don't think that the _solution_ to this issue that this PR proposes is an acceptable one. It follows suit with the behavior that Wikipedia does, but that behavior is arbitrary and suboptimal IMO. I have not been able to find any explanation for why do things that way (and this PR doesn't provide one), aside from someone else doing it.

Also, if we were to accept the solution, the implementation can be improved/done differently. But there's no point to discuss implementation details until we agree on the high level solution.

I've looked into the spec that @rtfb mentioned above very carefully, and also into how Go currently behaves. It turns out there's much inconsistency there.

The spec says one thing, but Go has internal code in `html/template` and `net/url` that encode fragment differently. Also, the way they handle escaping for query parameters and fragments is different too. The comments in `net/url` code for encoding fragment suggest something that is slightly counter to the spec (it's still valid, just overly aggressive escaping - meaning it escapes things that aren't required to be escaped according to the spec). It's also important to remember there is a difference between encoding the URL (which can require some url escaping), and encoding the URL inside HTML (which may require html escaping).

- https://github.com/golang/go/blob/go1.6.3/src/net/url/url.go#L141-L148

In addition to all the challenges in finding consensus there, not all browsers share the same implementation.

But most importantly, I have not been able to find any concrete issues that can arise from this, so it's very hard to "test" that any change I tried actually fixes something concrete.

My conclusion is the following. The factors above contribute to the difficulty of finding a correct, proper, actionable fix. This issue seems minor and mostly harmless. I think it's okay to defer until there are real problems that can be found and described (in the last few years, there haven't been other reports).

I will close the PR because I think we need to discuss the issue first, not the implementation details. If this is still an issue you'd like to see resolved, I will make the recommendation that you create an issue report and describe a reproducible failing scenario, a way in which the current behavior of blackfriday produces a visible problem that we can agree on. That way we can productively work towards finding an acceptable resolution. I'll also unassign myself from this because I don't want to prevent others from looking at this issue.

Thank you for reporting this and taking the time to create the PR, and apologies again for the delays!

PR #224, opened 2015-12-10
Fix for #180

### LINKED ISSUE #180: Newline required after last "reference style link" or it is not parsed.
```go
var test = "[Something][1]\n[1]: http://blah.com"

output := blackfriday.MarkdownCommon([]byte(test))
fmt.Printf(string(output))
```
produces:
```
<p>[Something][1]
[1]: <a href="http://blah.com">http://blah.com</a></p>
```

but
```go
var test = "[Something][1]\n[1]: http://blah.com\n"
```

correctly produces: ```<p><a href="http://blah.com">Something</a></p>```

The difference is the `\n` at the end of the test string. Is the Markdown text required to end with a new line? Unless I'm missing a good reason, could we make the parsing more lack, and not require the \n at the end?

rtfb:
Removed an early-out return statement that used to short circuit reference detection code if no trailing newline was present. Fixes #180.

Fixed some cosmetics along the way.

### DIFF
--- a/inline.go
+++ b/inline.go
@@ -275,8 +275,8 @@ func link(p *parser, out *bytes.Buffer, data []byte, offset int) int {
 		i++
 	}
 
-	// inline style link
 	switch {
+	// inline style link
 	case i < len(data) && data[i] == '(':
 		// skip initial whitespace
 		i++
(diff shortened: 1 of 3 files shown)


shurcooL:
LGTM.

Issue #228, opened 2015-12-22
Blackfriday adds extra blank line in code block within a list


cmbrandenburg:
Input:

```
* Foo

        bar

        qux
```

Expected output:

```
<ul>
<li><p>Foo</p>

<pre><code>bar

qux
</code></pre></li>
</ul>
```

Actual output:

```
<ul>
<li><p>Foo</p>

<pre><code>bar


qux
</code></pre></li>
</ul>
```

Notice the extra blank line between “bar” and “qux.”

shurcooL:
Are you using `blackfriday.MarkdownCommon` or something else?

cmbrandenburg:
I'm using Hugo, and I get the same result with `blackfriday-tool`. I can dig in to find out more details if you need them.

For what it's worth, the canonical Gruber Perl markdown program has the same bug.

shurcooL:
I can reproduce.

> For what it's worth, the canonical Gruber Perl markdown program has the same bug.

That's helpful to know, thanks.

Interestingly, it also happens if using fenced code block:

```
* Foo
    
    ```     
    bar

    qux
    ```
```

shurcooL:
Another test case:

```
* Foo
    
    ```     
    bar












    qux
    ```
```

That produces the same output as above.

rtfb:
I think I have a fix. I'll test it a bit more and post later today.

shurcooL:
Fixed via #229.

PR #229, opened 2015-12-26
Fix bug with gathering list item lines

rtfb:
This problem has been lurking since the very early days, apparently introduced with commit 583b3c5e. Strange that nobody stumbled upon it for so long.

Instead of swallowing an empty line and then reintroducing it back again in certain cases, collect the list item body in an unaltered form and let the recursive parsing call sort things out.

Fixes issue #228.

### DIFF
--- a/block.go
+++ b/block.go
@@ -1153,6 +1153,7 @@ gatherlines:
 		// and move on to the next line
 		if p.isEmpty(data[line:i]) > 0 {
 			containsBlankLine = true
+			raw.Write(data[line:i])
 			line = i
 			continue
 		}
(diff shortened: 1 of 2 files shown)


shurcooL:
This is an awesome fix for blackfriday! :clap:

I am surprised too. I suppose having more than 1 blank line in real world documents is pretty rare, but it wasn't correct even in that case (there was an extra blank line rendered). It's probably because most other Markdown renderers seem to have this bug too, so that might've been a deterrent for reporting this.

I'm not closely familiar with implementation details of `(*parser).listItem`, so all I can go on are that previous and new tests pass, and the code change looks reasonable with nothing that stands out to me. LGTM based on that.

PR #231, opened 2016-01-06
Use net/url package in isRelativeLink function

gbitten:
I made this patch to insert AbsolutePrefix in any relative link (ex: `!(image)[image.jpeg]`) . I know it breaks some blackfriday's tests, but maybe this is the correct way to work.

BTW, blackfriday is a great package.

### DIFF
--- a/html.go
+++ b/html.go
@@ -18,6 +18,7 @@ package blackfriday
 import (
 	"bytes"
 	"fmt"
+	"net/url"
 	"regexp"
 	"strconv"
 	"strings"
@@ -901,31 +902,10 @@ func doubleSpace(out *bytes.Buffer) {
 }
 
 func isRelativeLink(link []byte) (yes bool) {
-	// a tag begin with '#'
-	if link[0] == '#' {
-		return true
+	u, err := url.Parse(string(link[:]))
+	if err == nil {
+		return !u.IsAbs()
 	}
-
-	// link begin with '/' but not '//', the second maybe a protocol relative link
-	if len(link) >= 2 && link[0] == '/' && link[1] != '/' {
-		return true
-	}
-
-	// only the root '/'
-	if len(link) == 1 && link[0] == '/' {
-		return true
-	}
-
-	// current directory : begin with "./"
-	if bytes.HasPrefix(link, []byte("./")) {
-		return true
-	}
-
-	// parent directory : begin with "../"
-	if bytes.HasPrefix(link, []byte("../")) {
-		return true
-	}
-
 	return false
 }
 


html.go
```diff
@@ -901,31 +902,10 @@ func doubleSpace(out *bytes.Buffer) {
 }
 
 func isRelativeLink(link []byte) (yes bool) {
-	// a tag begin with '#'
-	if link[0] == '#' {
-		return true
+	u, err := url.Parse(string(link[:]))
```
shurcooL: What is the purpose of creating a new slice from `link` that is equivalent to `link` itself? I.e., why do `link[:]` instead of just `link`?

shurcooL:
> For example, the markdowm **`!()[img.jpg]`** would render  **``**.

`!()[img.jpg]` is not valid Markdown for an image. Did you mean ```?

If so, I can't reproduce:

```Go
package main

import (
	"os"

	"github.com/russross/blackfriday"
)

func main() {
	b := blackfriday.MarkdownBasic([]byte(`![](img.jpg)`))
	os.Stdout.Write(b)
}
```

That program outputs `<p></p>`, not ``.

Can you provide more details on what you're trying to do, and what's happening instead?

html.go
gbitten: I got this hint in stackoverflow (details [here](http://stackoverflow.com/questions/14230145/what-is-the-best-way-to-convert-byte-array-to-string)). The user ferhat elmas [commented](http://stackoverflow.com/questions/14230145/what-is-the-best-way-to-convert-byte-array-to-string#comment55642294_18615786) about this issue:
> a string is in effect a read-only slice of bytes. You can't convert byte array directly to string so first slice then string".

If you try a direct convert, you will get an error, like this example: 
http://play.golang.org/p/Wdr1_sOzMH

russross: You can convert directly from a byte slice to a string (note that in
general it has to make a copy of the contents to ensure that the resulting
string is immutable). The example you linked to uses a byte array. In that
case it is necessary to take a slice of the array, and then the slice can
be converted into a string.

gbitten:
> Did you mean ``?

Yes, that is it.

> Can you provide more details on what you're trying to do, and what's happening instead?

I am building an application where the pages are in one server and the images are in another server. So, I am using the `AbsolutePrefix` renderer parameter to easily referrer the images in another server.

This code shows how  I am using blackfriday:
```
package main

import (
	"os"
	"github.com/russross/blackfriday"
)

func main() {
	parameters := blackfriday.HtmlRendererParameters{
		AbsolutePrefix: "http://somehost",
	}
	renderer := blackfriday.HtmlRendererWithParameters(0, "", "", parameters)
	b := blackfriday.MarkdownOptions([]byte(`![](img.jpg)`), 
				renderer, blackfriday.Options{Extensions: 0})
	os.Stdout.Write(b)
}
```
Its output is:
```
<p><img src="img.jpg" alt=""></p>
```
But I would like to be:
```
<p><img src="http://somehost/img.jpg" alt=""></p>
```

shurcooL:
This intersects with #152.

/cc @rtfb The whole relative/absolute link logic is pretty messy, and it intersects poorly with `AbsolutePrefix` parameter. There are magical special cases that are not documented.

For example, in:

```Go
func (options *Html) maybeWriteAbsolutePrefix(out *bytes.Buffer, link []byte) {
	if options.parameters.AbsolutePrefix != "" && isRelativeLink(link) && link[0] != '.' {
		out.WriteString(options.parameters.AbsolutePrefix)
		if link[0] != '/' {
			out.WriteByte('/')
		}
	}
}
```

Why is there `&& link[0] != '.'`? It's very unclear what it's meant to achieve at a high level. I think that's not the right place, it should be factored into `isRelativeLink`.

According to blame, the change is from https://github.com/russross/blackfriday/pull/152, but that PR is not well documented.

I think to resolve these issues, we'd need to figure out the explicit rules for how all relative/absolute features/flags/options intersect and which ones take higher precedence. Until we do, people will keep making PRs to fit their needs that break/overwrite the changes was done by the previous person for their needs.

shurcooL:
That said, I think the change proposed in this PR is a good one. We should be using `url.URL.IsAbs`.

But we can't merge this PR as is because it breaks tests.

shurcooL:
Also, I think we should be using [`url.URL.ResolveReference`](https://godoc.org/net/url#URL.ResolveReference) to resolve relative URLs into absolute ones. It's better than:

```Go
out.WriteString(options.parameters.AbsolutePrefix)
if link[0] != '/' {
	out.WriteByte('/')
}
```

But that would be likely have to be a breaking API change as `AbsolutePrefix` should become `url.URL` rather than `string`. Of course, we could still parse the `string` on the fly, but then it becomes unclear what to do if it fails to parse.

PR #238, opened 2016-02-05
Fix adjacent lists rendering

### LINKED ISSUE #235: Adjacent ordered and unordered lists merged
Given the following markdown content:

```text
- First
- Last

1. First
2. Last
```

`blackfriday` renders

```text
<ul>
<li>First</li>

<li><p>Last</p></li>

<li><p>First</p></li>

<li><p>Last</p></li>
</ul>
```

I would expect it to render as:

```text
<ul>
<li>First</li>
<li>Last</li>
</ul>

<ol>
<li>First</li>
<li>Last</li>
</ol>
```

moorereason:
Documentation and tests are included.

Fixes #235

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,15 +180,15 @@ implements the following extensions:
 
     Terms must be separated from the previous definition by a blank line.
 
-*   **Strict Adjacent Lists**. Do not merge adjacent lists of differing types
-    into a single list.  This extension allows the following example to be two
-    distinct lists instead of merging them into a single unordered list.
+*   **Lax Lists**. Merge adjacent lists of differing types that are separated by
+    blank lines into a single list.  This extension allows the following example
+    to be rendered as a single unordered list instead of two distinct lists.
 
-        * Top
-        * Bottom
+        * Lions
+        * Tigers
 
-        1. First
-        2. Last
+        1. Bears
+        2. Goats
 
 *   **Footnotes**. A marker in the text that will become a superscript number;
     a footnote definition that will be placed in a list of footnotes at the
--- a/README.md
+++ b/README.md
@@ -180,6 +180,16 @@ implements the following extensions:
 
     Terms must be separated from the previous definition by a blank line.
 
+*   **Strict Adjacent Lists**. Do not merge adjacent lists of differing types
+    into a single list.  This extension allows the following example to be two
+    distinct lists instead of merging them into a single unordered list.
+
+        * Top
+        * Bottom
+
+        1. First
+        2. Last
+
 *   **Footnotes**. A marker in the text that will become a superscript number;
     a footnote definition that will be placed in a list of footnotes at the
     end of the document. A footnote looks like this:
--- a/block.go
+++ b/block.go
@@ -1174,9 +1174,9 @@ gatherlines:
 			p.dliPrefix(chunk) > 0:
 
 			if containsBlankLine {
-				// if strict lists mode is enabled, end the list if the type
+				// if lax lists mode is disabled, end the list if the type
 				// changed after a blank line.
-				if p.flags&EXTENSION_STRICT_LISTS != 0 &&
+				if p.flags&EXTENSION_LAX_LISTS == 0 &&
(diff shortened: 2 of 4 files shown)


shurcooL:
Just something to discuss, but is there a reason this extension can't be the default behavior?

Are there any valid situations where you might mix different types of lists (ordered vs unordered), not just forms, and expect them to be the same list?

moorereason:
I considered this a bug from the beginning, but based on the feedback I received on my original issue, I assumed that the default behavior was intentional.  That's why I created an optional extension.

It would be a breaking change, but I'm in favor of treating this like a bug and making this PR behavior the default.  If we did that, would we still want an extension at all?  I could change it to `EXTENSION_LAX_LISTS`, if needed.

shurcooL:
Please don't change the implementation yet. I just wanted to discuss the possibility.

Does anyone have thoughts on this?

rtfb:
The PR looks good, I'll gladly merge it when we settle down with the discussion.

I took a look at Markdown spec and it doesn't seem to require merging the lists. So looks like we can consider it as a bug and switch the default to a sane one. Having `EXTENSION_LAX_LISTS` would probably be a good idea in case there actually are people who depend on this behaviour.

Sorry for taking so long to chime in.

moorereason:
I've refactored this PR to add lax lists.  I can squash and rebase when needed.

shurcooL:
Given that this current behavior looks more like a bug than a useful feature anyone would actually rely on, having the flag be opt-out (`EXTENSION_LAX_LISTS`) rather than opt-in as originally proposed is great.

However, I want to point out that adding this flag is going to contribute to flag bloat for this library.

Here's an idea: let's not add the flag first. If there are many people who complain and ask for old behavior back with compelling reasons, then we can add the `EXTENSION_LAX_LISTS` flag then.

Otherwise we'll treat this as a simple bug fix. What do you think?

russross:
Does anyone know how CommonMark handles this? If the new behavior matches what they do, then that would be a strong reason to make it the default/only behavior.

moorereason:
@russross,
To be clear, this PR only changes the behavior of loose list items separated by a blank line (`1. One\n\n* Foo\n`).  Blackfriday would still allow you to be lazy (flagrantly so?) with list item types for tight lists (`1.One\n* Foo\n` would still be a single ordered list).  Blackfriday would end up handling these situations the same way Github does right now.  

In CommonMark a *list* is only a *list* if it's of the same type.  If the item type changes, it's a different list regardless of whether it's tight or loose.

My understanding of the markdown spec is that it's silent on this issue, but the [Markdown dingus](http://daringfireball.net/projects/markdown/dingus) matches the current blackfriday behavior.

I personally prefer the CommonMark approach, and this PR is a step closer to CommonMark.

shurcooL:
> If the new behavior matches what they do, then that would be a strong reason to make it the default/only behavior.

> In CommonMark a list is only a list if it's of the same type. If the item type changes, it's a different list regardless of whether it's tight or loose.

Sounds like that's the case.

:+1: from me to making the new behavior standard.

Another reason to do is because the new output looks a lot more like the textual input than the previous behavior.

shurcooL:
I think we should merge this since it's an improvement. Either as is, or without the flag since it seems to be very unlikely to be used by anyone, but will add to maintenance cost.

rtfb:
Agreed. I'll try to find some time in the upcoming days to clean the extension out and merge.

@moorereason, if you're still interested and can make the sane behavior the default, that will speed things up :-)

shurcooL:
It looks to be very easy. Just remove EXTENSION_LAX_LISTS and assume it's always not set in the rest of the code.

I can do it later today if @moorereason doesn't prefer to do it.

shurcooL:
Also, a suggestion. It might be better to do that on a new branch/PR and close this PR. Then we can use this PR as a reference in case we'd like to bring the `EXTENSION_LAX_LISTS` extension back sometime later.

moorereason:
I've submitted PR #261 to replace this PR.

shurcooL:
Awesome, thank you. I'll close this since the issue is now resolved, and we can look up what it takes to make it an extension if ever needed.

PR #257, opened 2016-04-28
Add EXTENSION_INLINE_CODE

emilkm:
When enabled, func codeSpan looks for ~languagename~ at the beginning of the code string

and outputs `"<code class=\"language-" + languagename + "\">"`
instead of just `"<code>"`
for the HTML renderer. (Latex ignores it.)

This enables prettier syntax highlighting of inline code with JavaScript libraries like prism.js

For example Markdown
~~~
`~php~$var = 0;`
~~~

would produce the following HTML

~~~
<code class="language-php">$var = 0;</code>
~~~

I know this is a long shot as it introduces non-standard markdown syntax, but maybe you would like to consider it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ _obj
 _test*
 markdown
 tags
+/.idea
+/blackfriday.iml
--- a/html.go
+++ b/html.go
@@ -482,8 +482,12 @@ func (options *Html) AutoLink(out *bytes.Buffer, link []byte, kind int) {
 	out.WriteString("</a>")
 }
 
-func (options *Html) CodeSpan(out *bytes.Buffer, text []byte) {
-	out.WriteString("<code>")
+func (options *Html) CodeSpan(out *bytes.Buffer, text []byte, lang string) {
+	if len(lang) > 0 {
+		out.WriteString("<code class=\"language-" + lang + "\">")
+	} else {
+		out.WriteString("<code>")
+	}
 	attrEscape(out, text)
 	out.WriteString("</code>")
 }
--- a/inline.go
+++ b/inline.go
@@ -147,7 +147,27 @@ func codeSpan(p *parser, out *bytes.Buffer, data []byte, offset int) int {
 
 	// render the code span
 	if fBegin != fEnd {
-		p.r.CodeSpan(out, data[fBegin:fEnd])
+		langStart := 0
+		langEnd := 0
+
+		if p.flags&EXTENSION_INLINE_CODE != 0 {
+			// look for ~languagename~ at the beginning of the code string
+			if data[fBegin] == '~' {
+				idx := fBegin + 1
+				langStart = idx
+
+				for idx < fEnd && data[idx] != '~' {
+					idx++
+				}
+
+				if idx < fEnd {
+					langEnd = idx
+					fBegin = idx + 1
+				}
+			}
+		}
+
+		p.r.CodeSpan(out, data[fBegin:fEnd], string(data[langStart:langEnd]))
 	}
 
 	return end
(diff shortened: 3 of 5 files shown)


rtfb:
`EXTENSION_INLINE_CODE` is a bit misleading. `EXTENSION_INLINE_CODE_LANG`?

What would this code do in a case like this?

    `~/.vimrc, ~/.bashrc`

Rant:
I am indeed less than happy to accept extensions to Markdown syntax. It's enough of a mess as it is. In general I'd like Blackfriday to be extensible enough so that people could implement such things as their own extensions, not pile things up here, but alas, it's not yet possible as of now.

emilkm:
Thanks for the response. Sorry about the trouble and if I am wasting your time. The change is obviously not all that well considered. If you would rather reject it, that's no problem. 

Your question about the name of the extension makes me think you are somewhat open to include something like this.

In that case I would be happy to 
* change the name to EXTENSION_INLINE_CODE_LANG
* come up with better delimiter than ~, or have an alternative (your input would be great)
* add a test
* do anything else you suggest

Either way, thanks for the great work on this. I am using it via hugo and it is doing a great job for me.

rtfb:
Oh no, you're not wasting anyone's time! The proposal is perfectly understandable, even if it proves to be difficult to implement. As to the alternatives, how about having parentheses immediately precede or follow the backticks? Like this:

    (go)`func foo() {}`
    `def foo(): pass`(python)

I think having the language specifier outside of backticks a safer approach, since the whole purpose of code snippets is to have them verbatim, so the less we interpret the contents, the better.

emilkm:
OK thanks for that. Agree about the language specifier being outside is safer. It even reads better, should the markdown contain one and the output was rendered without the extension enabled.

`<code>def foo(): pass</code>(python)` wouldn't look too bad.

I'd say that what might be easier for you could prove quite difficult for me :) having just started the other day with the go language.

I had a look and it seems that parentheses preceding the backticks would be harder. The `inline` function has already found the offset (the position of the opening backtick) when it calls the `codeSpan` function. It then checks to see what the `codeSpan` function has consumed. This would mean that the if the parentheses were following the backtics, the `codeSpan` can deal with it by looking past the closing backtick and check for the lang. Then simply set the cursor past the end of the lang if it finds it.

I'll have a go at it unless you think I am missing something obvious.

I'll close this request then.

Thanks.

shurcooL:
Hi @emilkm,

Are you already familiar with [Fenced Code Blocks](https://help.github.com/articles/creating-and-highlighting-code-blocks/)? Could that be used to satisfy your needs? With fenced code blocks, you can already specify the language.

For example:

    ```php
    $var = 0;
    ```

emilkm:
Hi,

Yes thanks for that. I am working on a documentation site with hugo and am using Fenced Code Blocks. I am using client side highlighting with Prism.js

Inline code works well most of the time, and is better when the content is only a path or just a parameter name. 

For example in https://gohugo.io/overview/configuration/ Configure Blackfriday rendering, it would be pointless to syntax highlight the flag names.

Sometime when referring to specific parts of longer code block, it is arguably nicer to have inline code highlighted.

For example https://laravel.com/docs/5.2/errors Ignoring Exceptions By Type

This is quite a small thing really, and I could happily live without it. In fact too much highlighting can be harder to read.

I do have it working for myself. It was fun compiling hugo and blackfriday and getting to know the go language.

I just realised that blackfiday is in the process of big changes for v2 and is probably just annoying to add little things like this at the moment.

I mad this change and automatic heading anchors (similar to github docs) work for me, and am as happy as Larry. Both blackfiday and hugo excellent and do everything else I needed, and even things I didn't know I needed until I found out you can do them :)

shurcooL:
I see. I just wanted to confirm you were aware of Fenced Code Blocks. Thanks for the explanation. :)

PR #258, opened 2016-04-29
Travis: Run gofmt -s, go vet, go test -race, add Go 1.6.

shurcooL:
Also make tip a fast-finish allowed failure. That way, if CI fails on tip due to a temporary issue with tip, it will not break build status. However, it's still possible to see tip build status by looking at CI details page.

Do not run go vet with Go 1.4 or older since it's not included in the standard library, and it's no longer available in external standard library.

Add godoc badge to README.md.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,30 @@
-# Travis CI (http://travis-ci.org/) is a continuous integration service for
-# open source projects. This file configures it to run unit tests for
-# blackfriday.
-
+sudo: false
 language: go
-
 go:
-    - 1.2
-    - 1.3
-    - 1.4
-    - 1.5
-
+  - 1.5.4
+  - 1.6.2
+  - tip
+matrix:
+  include:
+    - go: 1.2.2
+      script:
+        - go get -t -v ./...
+        - go test -v -race ./...
+    - go: 1.3.3
+      script:
+        - go get -t -v ./...
+        - go test -v -race ./...
+    - go: 1.4.3
+      script:
+        - go get -t -v ./...
+        - go test -v -race ./...
+  allow_failures:
+    - go: tip
+  fast_finish: true
 install:
-    - go get -d -t -v ./...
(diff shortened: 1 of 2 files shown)


shurcooL:
/cc @rtfb Are you able to review this?

If you have any questions about any parts of the Travis script, let me know, I know it well and I'm happy to explain anything that's unclear. It's quite close to the [standard Travis script](https://github.com/shurcooL/cmd/blob/29b3b215f24fed434922acf6333d605c9a7cc69a/gorepogen/main.go#L50-L66) I run on all my Go repos.

It's also very similar to the change I've made in https://github.com/google/go-github/pull/346.

rtfb:
Sure. Looks good. A couple questions:
- What does `diff -u <(echo -n) <(gofmt -d -s .)` do? Why not just `gofmt -d -s .`?
- Why specify the subreleases for 1.5 and 1.6 series? Wouldn't Travis pick up latest if we only specified major.minor?

shurcooL:
> What does `diff -u <(echo -n) <(gofmt -d -s .)` do? Why not just `gofmt -d -s .`?

`gofmt -d -s .` prints any deviations from gofmt style (including simplify option) to stdout. However, it always exits with 0, even if there are diffs. If someone submits a PR with invalid gofmt formatting, we want the Travis build to fail.

That's why I came up with the idea of doing a `diff -u` of an empty string against `gofmt -d -s .`. Non-zero difference will have a non-zero exit code.

> Why specify the subreleases for 1.5 and 1.6 series? Wouldn't Travis pick up latest if we only specified major.minor?

I'm happy to double check this (I'll make that change on another branch, then we can check Travis output), but last time I tried, specifying just 1.5 or 1.6 in Travis would actually use 1.5.0 and 1.6.0, not the latest version.

rtfb:
I see, thanks for explanations. (I wonder if Travis would consider this as a bug though...)

LGTM.

PR #262, opened 2016-05-04
Add EXTENSION_INLINE_CODE_LANG

emilkm:
Follow up from #257

Enables syntax highlighting of inline code with JavaScript libraries like prism.js

Adding a language specifier following the backticks in the markdown source

    `def foo(): pass`(python)

will render as (with extension enabled)

    <code class="language-python">def foo(): pass</code>

and (with extension NOT enabled)

    <code>def foo(): pass</code>(python)

Whitespace is allowed between the closing backtick and the language specifier. The following

    `def foo(): pass` (python)

will also render to

    <code class="language-python">def foo(): pass</code>


html.go
```diff
@@ -482,8 +482,12 @@ func (options *Html) AutoLink(out *bytes.Buffer, link []byte, kind int) {
 	out.WriteString("</a>")
 }
 
-func (options *Html) CodeSpan(out *bytes.Buffer, text []byte) {
-	out.WriteString("<code>")
+func (options *Html) CodeSpan(out *bytes.Buffer, text []byte, lang string) {
+	if len(lang) > 0 {
```
shurcooL: It's more idiomatic to check that a string is not empty by doing `if lang != "" {`.

markdown.go
```diff
@@ -179,7 +180,7 @@ type Renderer interface {
 
 	// Span-level callbacks
 	AutoLink(out *bytes.Buffer, link []byte, kind int)
-	CodeSpan(out *bytes.Buffer, text []byte)
+	CodeSpan(out *bytes.Buffer, text []byte, lang string)
```
shurcooL: I'll point out that this changes the public `Renderer` interface, that means it will make every existing implementation in the wild no longer import [`blackfriday.Renderer` interface](https://godoc.org/github.com/russross/blackfriday#Renderer). This doesn't mean it cannot be done, but it should not be taken lightly and discussed before doing.

inline.go
```diff
@@ -111,17 +111,17 @@ func emphasis(p *parser, out *bytes.Buffer, data []byte, offset int) int {
 
 func codeSpan(p *parser, out *bytes.Buffer, data []byte, offset int) int {
 	data = data[offset:]
-
+	datalen := len(data)
```
shurcooL: "data" and "len" are separate words, right? If so, this should be `dataLen` according to idiomatic Go style, see https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps.

shurcooL:
I left some inline Go style comments for your reference. However, this PR should first be reviewed (and accepted) at a high level, so you can put off making those changes until later if you wish. I just wanted to post them since they're general Go code review comments.

emilkm:
Thanks for the comments. I have done the changes.

I agree the change to Render interface should not be taken lightly.

rtfb:
Please see discussion at #288, I'd like to postpone taking PRs like this.

emilkm:
No problem.

PR #264, opened 2016-05-06
Fix definition lists that contain other lists

### LINKED ISSUE #263: Definition lists containing other lists are broken
PR #261 broke definition lists that contain other lists.  See the comments by @miekg on that PR.

Example test:

```
diff --git a/block_test.go b/block_test.go
index bb3a50b..f3618fe 100644
--- a/block_test.go
+++ b/block_test.go
@@ -960,6 +960,14 @@ func TestDefinitionList(t *testing.T) {
                        "<dd><p>Definition b</p></dd>\n" +
                        "</dl>\n" +
                        "\n<p>Text 2</p>\n",
+
+               "Term 1\n:   Definition a\n\n    Text 1\n\n    1. First\n    2. Second",
+               "<dl>\n" +
+                       "<dt>Term 1</dt>\n" +
+                       "<dd><p>Definition a</p>\n\n" +
+                       "<p>Text 1</p>\n\n" +
+                       "<ol>\n<li>First</li>\n<li>Second</li>\n</ol></dd>\n" +
+                       "</dl>\n",
        }
        doTestsBlock(t, tests, EXTENSION_DEFINITION_LISTS)
 }
```

moorereason:
Fixes #263

Ping @miekg @tianon

### DIFF
--- a/block.go
+++ b/block.go
@@ -1175,8 +1175,9 @@ gatherlines:
 
 			if containsBlankLine {
 				// end the list if the type changed after a blank line
-				if (*flags&LIST_TYPE_ORDERED != 0 && p.uliPrefix(chunk) > 0) ||
-					(*flags&LIST_TYPE_ORDERED == 0 && p.oliPrefix(chunk) > 0) {
+				if indent <= itemIndent &&
+					((*flags&LIST_TYPE_ORDERED != 0 && p.uliPrefix(chunk) > 0) ||
+						(*flags&LIST_TYPE_ORDERED == 0 && p.oliPrefix(chunk) > 0)) {
 
 					*flags |= LIST_ITEM_END_OF_LIST
 					break gatherlines
--- a/block_test.go
+++ b/block_test.go
@@ -960,6 +960,14 @@ func TestDefinitionList(t *testing.T) {
 			"<dd><p>Definition b</p></dd>\n" +
 			"</dl>\n" +
 			"\n<p>Text 2</p>\n",
+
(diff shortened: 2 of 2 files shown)


miekg:
yes, this passes my tests

miekg:
nice short fix btw :)

shurcooL:
I don't completely understand how it works, and I'm not familiar with definition lists (I never use them myself). The change looks reasonable and passes all existing tests, and fixes this one, so LGTM. /cc @rtfb

tianon:
Tested it with my example from https://github.com/russross/blackfriday/issues/235#issuecomment-217227395, and it works great. :+1:

block.go
```diff
@@ -1175,8 +1175,9 @@ gatherlines:
 
 			if containsBlankLine {
 				// end the list if the type changed after a blank line
```
shurcooL: Is this comment still up to date despite the code change below? Just checking.

moorereason: Should probably be updated.

rtfb: Should I wait for your change or should I merge right away?

moorereason:
@rtfb, you can probably merge this now, and we can improve upon it later.

My main concern is that this PR only solves half of the problem.  If you have a Def List that contains differing adjacent lists, the adjacent lists are merged into a single list (which is the problem that I was originally trying to fix).  I'm having a hard time understanding how to detect the nesting state to fix that problem and haven't had time to devote to fixing it.

Example test case:

```
"Term 1\n:   Definition a\n\n    Text 1\n\n    1. First\n    2. Second\n    * Foo\n    * Bar",
```

Issue #218, opened 2015-11-10
Discussion: initial work on v2


rtfb:
I have rebased my v2 experiments on top the latest master and pushed them to v2 branch for eyeballing.

Here's what I was doing and where I stand:
- All constants renamed from C_STYLE to IdiomaticGo (and plain ints to typed
  constants). The rename was mechanical so far, I didn't try to take a chance
  and look for opportunities to eliminate naming inconsistencies; but we
  probably should do that while we're breaking the compatibility. Suggestions
  welcome.
- Renderer interface overhaul. Eliminated callbacks from renderer methods.
  #120 suggested changing `*bytes.Buffer` to `io.Writer`, but I tried another
  approach: leaving all writing business to the renderer internals.
- #189 is probably fixed as side effect: no more callbacks, no more dead
  loops.
- When table of contents is enabled, it still behaves the old way: backups
  all the document in a temp buffer, truncates it, writes out TOC, writes back
  temp buffer. I don't see an easy way around it.

The next big thing I'm going to try is constructing an actual AST before rendering. This should allow to clean up our internal code substantially in some places and further simplify the renderer interface (and properly fix the TOC problem).

Comments welcome.

russross:
I haven't had a chance to look at it yet, but from your description I can say that this sounds like the right direction to me. I especially like the idea of moving to an AST where users can manipulate it freely and still use the normal HTML renderer when the are finished making changes.

The only thing I would add is this: don't feel like you need to wait for my approval when you are ready to commit to these large changes. When you think it is ready, you should be the one to make the call.

Thanks again for all of your work on this project!

egonelbre:
Some tips. Introduce a line-reader into the code. See https://github.com/loov/mark/blob/master/reader.go

This allows to write: 
* [main loop](https://github.com/loov/mark/blob/master/parse.go#L90)
* [fenced code](https://github.com/loov/mark/blob/master/parse.go#L267)
* [quoted blocks](https://github.com/loov/mark/blob/master/parse.go#L134)

*Of course, adjust reader to your own needs. e.g. [something smaller](https://github.com/egonelbre/mmark/blob/50dfad50e1706d9bd04c9bd35c0a369baf424449/markdown.go#L1002)*

client9:
Hi team,

This is a fantastic project.  

When using the API I bumped into a few oddities, that I'd love to see fixed up in a v2.  I'm happy to help or work on any of these idea.

##### Would like to know the position in the original markdown text.

[...]

PR #270, opened 2016-06-23
Add option LATEX_MATH, a.k.a. Mathjax math for HTML output

Ambrevar:
Make parsing of '$' and '$$' special. It makes it possible to typeset beautiful math using TeX-and-friends syntax. It can be rendered on an HTML page with 

```html
<script type="text/javascript" async="" src="http://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML"></script>
```  

Credits go to Yuntan for this patch, I just fixed the closing delimiter.

### DIFF
--- a/markdown.go
+++ b/markdown.go
@@ -46,6 +46,7 @@ const (
 	EXTENSION_AUTO_HEADER_IDS                        // Create the header ID from the text
 	EXTENSION_BACKSLASH_LINE_BREAK                   // translate trailing backslashes into line breaks
 	EXTENSION_DEFINITION_LISTS                       // render definition lists
+	EXTENSION_LATEX_MATH                             // latex inline and display math surrounded by '$' or '$$'
 
 	commonHtmlFlags = 0 |
 		HTML_USE_XHTML |
@@ -189,6 +190,7 @@ type Renderer interface {
 	TripleEmphasis(out *bytes.Buffer, text []byte)
 	StrikeThrough(out *bytes.Buffer, text []byte)
 	FootnoteRef(out *bytes.Buffer, ref []byte, id int)
+	Math(out *bytes.Buffer, equation []byte, inline bool)
 
 	// Low-level callbacks
 	Entity(out *bytes.Buffer, entity []byte)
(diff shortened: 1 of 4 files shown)


markdown.go
```diff
@@ -189,6 +190,7 @@ type Renderer interface {
 	TripleEmphasis(out *bytes.Buffer, text []byte)
 	StrikeThrough(out *bytes.Buffer, text []byte)
 	FootnoteRef(out *bytes.Buffer, ref []byte, id int)
+	Math(out *bytes.Buffer, equation []byte, inline bool)
```
shurcooL: I don't think we can accept a change that changes the `Renderer` interface at this stage, as it would invalidate all existing `blackfriday.Renderer` implementations.

Ambrevar:
Good point.

This, however, highlights what might be a serious design issue with Blackfriday: the parser and the renderer interface are set up upside down which offers poor interfacability.

Currently, renderers are merely translators, they have no control over the rendering outline which is hard-coded in `markdown.go:secondPass()`.
It severely limits rendering possibilities. For instance, the LaTeX title should be set in the preamble, not in the body, but the Renderer interface does not allow for that.

I believe that the interface is set up the wrong way: The Parser should be a structure holding methods like TitleBlock() which would return the content of the title, and so on. The Renderer interface should have one method only, namely 'Render', which would call on the various parser methods, like `secondpass()` does. As a consequence, extending the parser would not break backward compatibility with existing renderers.

For now, my proposition does break backward compatibility with existing third-party renderers. I believe a fork could make future versions much stronger in regard to backward-compatibility.

shurcooL:
> This, however, highlights what might be a serious design issue with Blackfriday: the parser and the renderer interface are set up upside down which offers poor interfacability.

That is true. There is a blackfriday v2 which attempts to address those issues with the interface. Have you looked at it? Would you consider checking how your change could fit on top of v2 instead?

https://github.com/russross/blackfriday/issues/218

According to @rtfb it is mostly ready and simply needs to be used and tested by people.

Ambrevar:
I missed that, thanks for the hint! I don't have time at the moment, I'll give it a look as soon as I can.

Ambrevar:
I've sent a PR for v2, see #271.

Issue #256, opened 2016-04-27
any plan to support mathjax


xiexinyls:
Nowadays a lot of people include math in markdown with delimit $, $$, \(\), \[\], which blackfriday doesn't support.
I have seen extension or plugins in the markdown converter in Python supporting skipping math expression. This usually is implemented by replace math expression with some special token first and run the native markdown converter. After that, they replace the tokens with the actual math expression back. Is there any plan to support that in blackfriday?

Very looking forward to that feature!

Depado:
I'd love to see that too :)

Ambrevar:
See #271.

Issue #220, opened 2015-11-19
using BlackFriday with math


amlwwalker:
I am rendering maths with MathJax library, and for some reason `\\` is escaped to `\`, when I put `\\\` that gets escaped to `\\` whcih is what I want. Where would this be happening, and why?

Thanks

shurcooL:
What about if you put `\\\\`, does that get correctly escaped to `\\`?

PR #271, opened 2016-06-30
Add Math extension

Ambrevar:
It enables direct math support in LaTeX, as well as HTML with MathJax.

I believe this feature to be a popular demand. See #256 and #220.

This is a request for comments, as I am quite new to Blackfriday.
I haven't written many tests yet, I'm waiting for your comments.

The main point of contention would be the math delimiter, being '$' and '$$' at the moment. This obviously creates a conflict with a regular use of the dollar signe. Both MathJax and LaTeX use '\(' and '\['. I am not sure how we can use those in Markdown because of the conflict with the 'escape' rules.

### DIFF
--- a/block.go
+++ b/block.go
@@ -194,6 +194,13 @@ func (p *parser) block(data []byte) {
 			}
 		}
 
+		if p.flags&LaTeXMath != 0 {
+			if i := p.latexMath(data); i > 0 {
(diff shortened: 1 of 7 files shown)


Ambrevar:
I've inadvertently deleted the branch... Recreating the PR in #272.

PR #273, opened 2016-07-01
v2: Fix TOC generation

Ambrevar:
Track levels properly.
Support headers with multiple/nested children.
Can jump for more than 1 level at once.

This path highlights a "weakness" of the current implementation: if the renderer wants to use "all the text of the node and its children", it needs some work... I've writtent a deep copy function for nodes that could help a lot here, so we should consider exporting it.

Any thoughts on this?

### DIFF
--- a/block_test.go
+++ b/block_test.go
@@ -1526,6 +1526,71 @@ func TestTOC(t *testing.T) {
 <h1 id="toc_2">Title2</h1>
 `,
 
+		"## Subtitle\n\n# Title",
+		`<nav>
+
+<ul>
+<li>
+<ul>
+<li><a href="#toc_0">Subtitle</a></li>
+</ul></li>
+
+<li><a href="#toc_1">Title</a></li>
+</ul>
+
+</nav>
+
+<h2 id="toc_0">Subtitle</h2>
+
+<h1 id="toc_1">Title</h1>
+`,
+
+		"# Title 1\n\n## Subtitle 1\n\n### Subsubtitle 1\n\n# Title 2\n\n### Subsubtitle 2",
+		`<nav>
+
+<ul>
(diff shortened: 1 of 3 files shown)


rtfb:
Hey @Ambrevar,

Sorry for taking so long to review this PR. Overall it looks good to me, my only concerns are with the `deepCopy`:

1. Copying `content` is definitely not needed. This field carries block content from the block parser to inline and is set to `nil` after parsing is complete. So by the time TOC is constructed it will always be `nil`.
2. Copying `Literal` is also unnecessary for the task at hand -- a reference to the original slice does the same job, while avoiding allocations and copies. But that would spoil the "deepness" of the `deepCopy`, of course.
3. However, I don't think there's much need in exporting `deepCopy`. And if we don't export it, maybe "semi-deepness" is not really a problem.
4. I'm still thinking whether it would be possible to get away without deep copying anything at all. Conceptually I don't see why a simple reference to the `Header` subtree wouldn't work. But when I tried it, it obviously didn't and I haven't had a chance to debug thoroughly what's going on.

Ambrevar:
1. Ack.
2. Ack.
3. Ack.
4. No, there is no way around a semi-deep copy because of the way the AST is walked and the way the TOC is contructed: if you append the header directly the TOC under construction, either you update the header parents, in which case you'll break the text, or you don't, in which case when the TOC will be rendered, the walk will follow the header parent which is the rest of the body, it will not go back to the TOC, and thus fail after the first entry.

Solutions I see here:

- This commit using the suggestions you mentionned in 1 and 2.
- Move the TOC generation to the renderer and instead of hacking the AST, we render as we parse. I think this solution would be much saner.

rtfb:
Superseded by #294, closing.

PR #274, opened 2016-07-01
v2: Implicit interface and Stringer

Ambrevar:


node.go
```diff
@@ -310,6 +310,10 @@ func (nw *NodeWalker) resumeAt(node *Node, entering bool) (*Node, bool) {
 	return nw.next()
 }
 
+func (ast *Node) String() string {
+	return dumpString(ast)
+}
```
shurcooL: Do I understand correctly that this would dump the entire AST of the document recursively? Wouldn't that be a very large output?

I'm not sure if that's a great fit for `fmt.Stringer` output. Usually it's meant to be short descriptions that can fit into `fmt.Errorf("something went wrong with %v in FooBar: %v", htmlRenderer, err)` messages, etc.

Do you want this functionality for debugging?

Ambrevar:
It outputs the node and its children, not necessarily the whole document. The output may not be that large.
It would be helpful for debugging, since some fields are not exported and thus not available for printing.

Another solution would be to output the current node only and leave the recursive part to a 'Dump' function.

shurcooL:
> Another solution would be to output the current node only and leave the recursive part to a 'Dump' function.

I'm not quite sure about this, but it feels like using `fmt.Stringer` interface for debugging in a public library API might not be a great fit.

I can see it go either way though.

I am leaning towards having dedicated Dump functions though. Let's compare with go/ast:

```Go
astFileNode, err := parser.ParseFile(...)

// Prints a few lines.
fmt.Println(astFileNode)

// Prints the whole tree.
ast.Fprint(os.Stdout, fset, astFileNode, nil)
```

I worry that relying on `String()` to be recursive for debugging would be too inflexible, since there can only be one implementation of it.

rtfb:
I agree with @shurcooL on this one. Let's have a dedicated `Dump` function for listing all the fields of a node. And then maybe rename `dumpString` to `DumpTree` for the one that recurses. Although I can only guess if it's of any use to expose such debugging funcs to the public.

Ambrevar:
Agreed on leaving the recursive printing to a separate function, however I would still make a stringer out of Node, just like go/ast does, containing the most important fields. (Type and Literal?)

I believe the export to be useful to people writing Renderers since they work with the AST quite a bit.

rtfb:
Oh yes, printing a select few fields in `String` is fine. `Type` and `Literal` sounds good, but I think it would be nice to ellipsize the literal if it's longer than, say, 32 characters.

Ambrevar:
Ack.

Also please note that this PR contains another commit making HTML an implicit Renderer interface. Quite important I believe.

rtfb:
Yup, noted. No objections to this one :+1:

shurcooL:
> Also please note that this PR contains another commit making HTML an implicit Renderer interface. Quite important I believe.

Yep, I've seen that, sounds good.

It aligns with something @davecheney [said recently](https://twitter.com/davecheney/status/747953959529119744):

> Return a public type rather than an interface.

html.go
```diff
@@ -124,7 +124,7 @@ func (w *HTMLWriter) Newline() {
 
 // NewHTMLRenderer creates and configures an HTMLRenderer object, which
 // satisfies the Renderer interface.
-func NewHTMLRenderer(params HTMLRendererParameters) Renderer {
+func NewHTMLRenderer(params HTMLRendererParameters) *HTMLRenderer {
```
rtfb: And while you're at it, could you also please apply the same fix to `NewLatexRenderer`?

Ambrevar: Sure.

Talking about the LaTeX renderer, it's currently a stub. Is there any plan for its implementation? If not, I'd be happy to work on it. I've already commited some changes to my fork (but that's for v1).

rtfb: Latex renderer is currently a bit of an orphaned child. It was decided that different renderers are not a priority for Blackfriday, up to the point where people are welcome to extract it to a separate repo and maintain it there.

Ambrevar: What shall we decide for the Latex renderer? In its current state it is obviously useless. Maintaining 2+ renderers is good practice to make sure the parser interface is not too biased toward one implementation of a renderer.
Thus I would suggest someone maintains a LaTeX renderer that is on par with the HTML renderer, either in this repository or in another.

If you want to drop it from this repo, I can implement my own and reference it here.

shurcooL: I'm in favor of dropping it from this repo. That way we can focus more on Markdown parsing here.

PR #275, opened 2016-07-01
v2: Fix empty path to CSS

Ambrevar:


html.go
```diff
@@ -748,7 +748,7 @@ func (r *HTMLRenderer) writeDocumentHeader(w *bytes.Buffer, sr *SPRenderer) {
 	w.WriteString(">\n")
 	if r.CSS != "" {
 		w.WriteString("  <link rel=\"stylesheet\" type=\"text/css\" href=\"")
-		attrEscape([]byte(r.CSS))
+		w.Write(attrEscape([]byte(r.CSS)))
```
shurcooL: It looks like most other code uses `esc` instead of calling `attrEscape` directly.

Also, both `esc` and `escCode` take 2 parameters, but the second parameter is never used. That's strange. /cc @rtfb

shurcooL:
LGTM but perhaps @rtfb will have style suggestions.

html.go
rtfb: Yeah, all of that (lack of `Write`, `s/esc/attrEscape` and the unused parameter) are wrong and the result of hectic hacking. I'm pulling this in and will fix the rest right away.

PR #280, opened 2016-07-15
Fix fenced code block rendering with content resembling references.

shurcooL:
See commit messages for details. Highly recommended to review commit-by-commit rather than entire diff at once.

### DIFF
--- a/block.go
+++ b/block.go
@@ -515,7 +515,7 @@ func (p *parser) htmlFindEnd(tag string, data []byte) int {
 	return i + skip
 }
 
-func (p *parser) isEmpty(data []byte) int {
+func (*parser) isEmpty(data []byte) int {
 	// it is okay to call isEmpty on an empty buffer
 	if len(data) == 0 {
 		return 0
@@ -530,7 +530,7 @@ func (p *parser) isEmpty(data []byte) int {
 	return i + 1
 }
 
-func (p *parser) isHRule(data []byte) bool {
+func (*parser) isHRule(data []byte) bool {
 	i := 0
 
 	// skip up to three spaces
--- a/block.go
+++ b/block.go
@@ -102,7 +102,7 @@ func (p *parser) block(out *bytes.Buffer, data []byte) {
 		// }
 		// ```
 		if p.flags&EXTENSION_FENCED_CODE != 0 {
-			if i := p.fencedCode(out, data, true); i > 0 {
+			if i := p.fencedCodeBlock(out, data, true); i > 0 {
 				data = data[i:]
 				continue
 			}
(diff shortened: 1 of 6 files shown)


markdown_test.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func TestDocument(t *testing.T) {
+	var tests = []string{
+		// Empty document.
+		"",
+		"",
+
+		" ",
+		"",
```
shurcooL: We previously had no regression tests for empty document, and I didn't know what was the expected/correct behavior for it.

So I've added that here. The current behavior is they both produce empty output (rather than a single newline, for example). Are we okay with that? It's okay by me.

rtfb: Can't imagine a use for producing anything else for this input. Fine by me too.

shurcooL:
FWIW, an alternative much simpler solution I considered, but decided not to use was this:

```diff
diff --git a/markdown.go b/markdown.go
index 58ba68d..eafead0 100644
--- a/markdown.go
+++ b/markdown.go
@@ -386,11 +386,25 @@ func MarkdownOptions(input []byte, renderer Renderer, opts Options) []byte {
 }
 
 // first pass:
+// - normalize (or add if missing) newline at end of non-empty? document
 // - normalize newlines
 // - extract references (outside of fenced code blocks)
 // - expand tabs (outside of fenced code blocks)
 // - copy everything else
 func firstPass(p *parser, input []byte) []byte {
+	// Normalize (or add if missing) newline at end of non-empty document.
+	// This is needed so that fenced code block detection works correctly during
+	// firstPass, since it's required for a fenced code block to end with newline
+	// to be detected.
+	/*switch {
+	case len(input) > 0 && input[len(input)-1] == '\n':
+		// Do nothing.
+	case len(input) > 0 && input[len(input)-1] == '\r':
+		input[len(input)-1] = '\n'
+	default:
+		input = append(input, '\n')
+	}*/
+
 	var out bytes.Buffer
 	tabSize := TAB_SIZE_DEFAULT
 	if p.flags&EXTENSION_TAB_SIZE_EIGHT != 0 {
```

The problem was that it caused an extra allocation/copy of the input document, and it modified input, which I didn't think was acceptable.

rtfb:
Thanks, @shurcooL! :-)

block.go
```diff
@@ -641,22 +645,26 @@ func (p *parser) isFencedCode(data []byte, syntax **string, oldmarker string) (s
...
 
-func (p *parser) fencedCode(out *bytes.Buffer, data []byte, doRender bool) int {
-	var lang *string
-	beg, marker := p.isFencedCode(data, &lang, "")
+// fencedCodeBlock returns the end index if data contains a fenced code block at the beginning,
+// or 0 otherwise. It writes to out if doRender is true, otherwise it has no side effects.
+// If doRender is true, a final newline is mandatory to recognize the fenced code block.
+func (p *parser) fencedCodeBlock(out *bytes.Buffer, data []byte, doRender bool) int {
+	var syntax string
+	beg, marker := isFenceLine(data, &syntax, "", true)
```
shurcooL: This should've been `false` for newline optional, it's mandatory here.

PR #281, opened 2016-07-15
Make newline mandatory for opening fence line.

shurcooL:
This was an unintended typo/mistake in #280.

This is stricter, and it's fine. The opening fence line will always need to have a newline.

Add another test for `isFenceLine`.


block.go
```diff
@@ -664,7 +664,7 @@ func isFenceLine(data []byte, syntax *string, oldmarker string, newlineOptional
 // If doRender is true, a final newline is mandatory to recognize the fenced code block.
 func (p *parser) fencedCodeBlock(out *bytes.Buffer, data []byte, doRender bool) int {
 	var syntax string
-	beg, marker := isFenceLine(data, &syntax, "", true)
+	beg, marker := isFenceLine(data, &syntax, "", false)
```
shurcooL: The boolean parameter is not the most readable; any suggestions for how it could be improved?

rtfb:
LGTM

PR #283, opened 2016-07-25
Update README to reflect security

NerdsvilleCEO:


russross:
I think this is all covered pretty thoroughly in the README already. If someone is concerned about security they should be reading beyond the executive summary anyway.

shurcooL:
There is a difference between being safe to user input (which is what blackfriday does), and sanitizing out parts of user input from the rendered output HTML (which is what an HTML sanitizer like bluemonday does).

This PR currently replaces one for the other, but they're different things.

NerdsvilleCEO:
Makes sense, my only concern is that it states that it is safe to provide user-supplied data to this library as it is "paranoid about the input", it just seems like a slight contradiction to the goal of this project in my opinion :)

shurcooL:
> my only concern is that it states that it is safe to provide user-supplied data to this library as it is "paranoid about the input",

It is safe to do that. What's not safe is to use the output (the output is the _rendered_ version of user-supplied data) without sanitizing it. Do you think that part is not clear?

NerdsvilleCEO:
I just thought that it could potentially be misleading, it was my misunderstanding :)

PR #284, opened 2016-07-27
Fix most of lint errors on v2

rtfb:
Since on of the things lint complains about are docstrings, documentation got somewhat improved.

### DIFF
--- a/markdown.go
+++ b/markdown.go
@@ -492,7 +492,7 @@ func (p *parser) parseRefsToAST() {
 		return
 	}
 	p.tip = p.doc
-	finalizeHtmlBlock(p.addBlock(HTMLBlock, []byte(`<div class="footnotes">`)))
+	finalizeHTMLBlock(p.addBlock(HTMLBlock, []byte(`<div class="footnotes">`)))
 	p.addBlock(HorizontalRule, nil)
 	block := p.addBlock(List, nil)
 	block.ListFlags = ListTypeOrdered
@@ -518,7 +518,7 @@ func (p *parser) parseRefsToAST() {
 	above := block.Parent
 	finalizeList(block)
 	p.tip = above
-	finalizeHtmlBlock(p.addBlock(HTMLBlock, []byte("</div>")))
+	finalizeHTMLBlock(p.addBlock(HTMLBlock, []byte("</div>")))
 	block.Walk(func(node *Node, entering bool) WalkStatus {
 		if node.Type == Paragraph || node.Type == Header {
 			p.currBlock = node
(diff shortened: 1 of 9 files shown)


markdown.go
```diff
@@ -58,18 +61,21 @@ const (
...
 		SmartypantsFractions | SmartypantsDashes | SmartypantsLatexDashes
 )
 
+// DefaultOptions is a convenience variable with all the options that are
+// enabled by default.
 var DefaultOptions = Options{
 	Extensions: CommonExtensions,
 }
 
+// TODO: this should probably be unexported. Or moved to node.go
```
shurcooL: Are you planning to address this TODO as part of this PR, or in a future change?

markdown.go
```diff
@@ -312,29 +320,21 @@ func MarkdownBasic(input []byte) []byte {
...
 // * Tables
-//
 // * Fenced code blocks
-//
 // * Autolinking
-//
 // * Strikethrough support
-//
 // * Strict header parsing
-//
```
shurcooL: I think these blank lines are needed to keep the bullet points as separate paragraphs.

https://godoc.org/gopkg.in/russross/blackfriday.v2#MarkdownCommon

Otherwise, all of them will be merged into a single paragraph, which I'm sure is not what you intended:

image

smartypants.go
```diff
@@ -398,13 +400,14 @@ func NewSmartypantsRenderer(flags Extensions) *SPRenderer {
 	return &r
 }
 
-func (sr *SPRenderer) Process(text []byte) []byte {
+// Process is the entry point of the Smartypants renderer.
+func (smrt *SPRenderer) Process(text []byte) []byte {
```
shurcooL: I don't have anything against it, but why `sr` -> `smrt` change? What does `smrt` stand for? Was this a golint issue? Just wondering.

shurcooL:
These are such great changes, thank you for doing them!

See comments, but otherwise LGTM.

markdown.go
rtfb: Future. It needs a bit of investigation to figure out whether it's a private field or should be a part of LinkData at node.go.

markdown.go
rtfb: Ahh, I see. That is not what I intended indeed. Thanks!

smartypants.go
rtfb: Yeah, golint complained about inconsistent receiver naming (all the rest funcs have `smrt`). I like `sr` better, but chose less intrusive change for this commit.

shurcooL: Ah, I see. In that case I'd recommend going with `sr` or even just `r` since `smrt` is a very long receiver name and I don't think we should keep it only because it's already used in more places. No harm in a bigger diff if it makes the code better.

rtfb: Yup, will do, hopefully later today.

PR #285, opened 2016-07-28
Unpublish and rename LinkType constants

rtfb:
The constants are only used in the parsing phase, they are not recorded
in the AST directly, so make them private. Improve their names along the
way. Fix tagLength to return two values instead of taking an output
parameter.

### DIFF
--- a/inline.go
+++ b/inline.go
@@ -638,13 +638,12 @@ func stripMailto(link []byte) []byte {
 	}
 }
 
-// autoLinkType specifies a kind of autolink that gets detected.
-type autoLinkType int
+// autolinkType specifies a kind of autolink that gets detected.
+type autolinkType int
 
 // These are the possible flag values for the autolink renderer.
-// Only a single one of these values will be used; they are not ORed together.
 const (
-	notAutolink autoLinkType = iota
+	notAutolink autolinkType = iota
 	normalAutolink
 	emailAutolink
 )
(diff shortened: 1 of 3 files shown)


inline.go
```diff
@@ -638,23 +638,33 @@ func stripMailto(link []byte) []byte {
 	}
 }
 
+// autoLinkType specifies a kind of autolink that gets detected.
+type autoLinkType int
+
+// These are the possible flag values for the autolink renderer.
+// Only a single one of these values will be used; they are not ORed together.
```
shurcooL:
```Go
// Only a single one of these values will be used; they are not ORed together.
```

I would leave this sentence out. That should be considered as the default for enums. Saying it explicitly makes sense if it weren't the case.

inline.go
```diff
@@ -638,23 +638,33 @@ func stripMailto(link []byte) []byte {
...
+const (
+	notAutolink autoLinkType = iota
```
shurcooL: Seeing an inconsistency here. Is it one word like `Autolink`/`autolink` or two words like `AutoLink`/`autoLink`?

I'd suggest one word like `Autolink`/`autolink` and use that everywhere.

shurcooL:
Less exported symbols? ✅
Still gets the job done? ✅
Result: 🎉 🎊

See comments, otherwise LGTM. This is a nice improvement!

PR #286, opened 2016-07-29
Repeat a run of reference tests under benchmark

rtfb:
Currently v2 is ~2.7x slower than v1. I have a bunch of optimizations of varying stages of hackiness on my box that (currently) bring it down to ~1.3x slower. I'm planning to push those changes to a separate optimizations branch bit by bit. So I want to get this benchmark landed on v2 before I branch off.

### DIFF
--- a/ref_test.go
+++ b/ref_test.go
@@ -80,7 +80,7 @@ func TestReference_EXTENSION_NO_EMPTY_LINE_BEFORE_BLOCK(t *testing.T) {
 var benchResultAnchor string
 
 func BenchmarkReference(b *testing.B) {
-	params := TestParams{Options: Options{Extensions: NoExtensions}}
+	params := TestParams{Options: Options{Extensions: CommonExtensions}}
 	files := []string{
 		"Amps and angle encoding",
 		"Auto links",
@@ -115,6 +115,7 @@ func BenchmarkReference(b *testing.B) {
 		}
 		tests = append(tests, string(inputBytes))
 	}
+	b.ResetTimer()
 	for n := 0; n < b.N; n++ {
 		for _, test := range tests {
 			benchResultAnchor = runMarkdown(test, params)
--- a/ref_test.go
+++ b/ref_test.go
@@ -13,7 +13,11 @@
 
 package blackfriday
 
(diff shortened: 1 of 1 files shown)


ref_test.go
```diff
@@ -70,3 +74,50 @@ func TestReference_EXTENSION_NO_EMPTY_LINE_BEFORE_BLOCK(t *testing.T) {
...
+	var tests []string
+	for _, basename := range files {
+		filename := filepath.Join("testdata", basename+".text")
+		inputBytes, err := ioutil.ReadFile(filename)
+		if err != nil {
+			b.Errorf("Couldn't open '%s', error: %v\n", filename, err)
+			continue
+		}
+		tests = append(tests, string(inputBytes))
+	}
```
shurcooL: I'll suggest adding a `b.ResetTimer()` line here. That way the benchmark will be less influenced by the file reading time.

ref_test.go
```diff
@@ -70,3 +74,50 @@ func TestReference_EXTENSION_NO_EMPTY_LINE_BEFORE_BLOCK(t *testing.T) {
 	}
 	doTestsReference(t, files, NoEmptyLineBeforeBlock)
 }
+
+// benchResultAnchor is an anchor variable to store the result of a benchmarked
+// code so that compiler could never optimize away the call to runMarkdown()
+var benchResultAnchor string
+
+func BenchmarkReference(b *testing.B) {
+	params := TestParams{Options: Options{Extensions: NoExtensions}}
```
shurcooL: How representative is this configuration of real world usage? Does using "common" flags have a significant effect on the performance?

shurcooL:
See comments. LGTM.

ref_test.go
rtfb: Ooh, that's a serious oversight. Adding common flags increased both version's run times considerably. And there seem to be unnecessary time eaters in those code paths. Thanks for spotting this!

ref_test.go
rtfb: I added it, but, a bit surprisingly, it didn't do much difference, almost undetectable, under 1%.

PR #293, opened 2016-08-09
v2: Export tree manipulation functions

Ambrevar:
Those functions are very similar to [Go's HTML package](https://godoc.org/golang.org/x/net/html), but the arguments differ: Go uses the parent while BF doesn't. I don't see a need for the parent, I find BF's way simpler. But Go developers aren't stupid, so there must be a good reason. If that reason happened to be relevant to Blackfriday, then we should make the same changes.

Comments?

### DIFF
--- a/block.go
+++ b/block.go
@@ -749,7 +749,7 @@ func (p *parser) table(data []byte) int {
 	i, columns := p.tableHeader(data)
 	if i == 0 {
 		p.tip = table.Parent
-		table.unlink()
+		table.Unlink()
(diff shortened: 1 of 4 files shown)


rtfb:
Now when they're public, could you also add doc comments to them, so that golint wouldn't complain (and the godocs page looked nice)?

Ambrevar:
Yep, forgot about that.

Ambrevar:
Done.

PR #294, opened 2016-08-09
v2: Move TOC generation to the HTML Renderer

Ambrevar:
Compare with #273 : this implementation is shorter, simpler, faster, easier to understand, and does not get in the way of other renderers.

### DIFF
--- a/block_test.go
+++ b/block_test.go
@@ -1532,6 +1532,71 @@ func TestTOC(t *testing.T) {
 <h1 id="toc_2">Title2</h1>
 `,
 
+		"## Subtitle\n\n# Title",
+		`<nav>
+
+<ul>
+<li>
+<ul>
+<li><a href="#toc_0">Subtitle</a></li>
+</ul></li>
+
+<li><a href="#toc_1">Title</a></li>
+</ul>
+
+</nav>
+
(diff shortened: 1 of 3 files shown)


rtfb:
This is indeed much more elegant! Please fix the details to pass the tests and let's go with this approach. Neat!

Should I already close #273?

Ambrevar:
Oopsy! Looks like I commited a bit too fast! :p Please close #273, I'll fix this one asap.

Ambrevar:
Fixed.

Ambrevar:
While we are it: shouldn't we export the writeDocumentHeader, writeDocumentFooter and writeTOC functions? I'm writing an HTML generator at the moment which is similar to BF's default with just a few changes to Render(). Exporting the write* functions would save a lot of copying.

As a general case, I think it would be good practice to export the supporting functions of the Render() function.

rtfb:
Since v2 started off having simplification of overly complex API as a primary goal, I was trying to expose as little things as functionally possible. I find the `writeDocument{Header,Footer}` funcs very peculiar to the `CompletePage` extension and can't really imagine much use in their exact reuse in other renderers.

Issue #253, opened 2016-04-05
commonHtmlFlags broken


c4s4:
commonHtmlFlags = 0 |
        HTML_USE_XHTML |
        HTML_USE_SMARTYPANTS |
        HTML_SMARTYPANTS_FRACTIONS |
        HTML_SMARTYPANTS_DASHES |
        HTML_SMARTYPANTS_LATEX_DASHES

Implies that:
- Output is XHTML (this XML).
- Simple and double quotes are replaced with & ldquo; and &lsquo; which are NOT valid XML entities.

Thus, I suggest that commonHtmlFlags are just HTML_USE_XHTML.

PR #295, opened 2016-08-09
v2: Factor Smartypants to HTML

Ambrevar:
Smartypants is HTML-specific.
There is no need to run Smartypants from `Render()`.
This simplifies extensions built upon the HTML renderer.

### DIFF
--- a/helpers_test.go
+++ b/helpers_test.go
@@ -106,7 +106,7 @@ func doLinkTestsInline(t *testing.T, tests []string) {
 		HTMLRendererParameters: params,
 	})
 	doTestsInlineParam(t, transformTests, TestParams{
-		HTMLFlags:              CommonHTMLFlags,
+		HTMLFlags:              UseXHTML,
 		HTMLRendererParameters: params,
 	})
 }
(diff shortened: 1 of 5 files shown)


rtfb:
Neat. But please rebase, it seems to have conflicted with the TOC changes that I just merged.

Fun fact: these changes should be beneficial to performance; I had similar changes in my performance tuning branch that still awaits cleanup.

Ambrevar:
The perf boost is actually very expectable: previously, the tree was walked twice, once for smarty pants, once for the normal rendering. This commits completely removes the SP walk.

Yet it could be even faster to move the condition out of the rendering loop: then we need 2 (almost identical) RenderNode() functions, which is very unclean, and I don't think the (probably limited) perf increase is worth the trouble.

Ambrevar:
Rebased.

Ambrevar:
Just noticed that this fixes #253 , doesn't it?

rtfb:
No, it doesn't fix #253, because `commonHTMLFlags` still contain both XHTML and Smartypants and they are contradictory.

But it is an improvement, so in it goes and thanks a lot! :-)

PR #297, opened 2016-08-09
v2: Automatically set title from the titleblock if it is unspecified

Ambrevar:
Not bulletproof, but probably better than nothing.

### DIFF
--- a/html.go
+++ b/html.go
@@ -752,6 +752,27 @@ func (r *HTMLRenderer) writeDocumentFooter(w *bytes.Buffer) {
 	w.WriteString("\n</body>\n</html>\n")
 }
 
+// expandTitle sets to title to the first line of the title block, in Markdown.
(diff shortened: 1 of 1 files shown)


rtfb:
But... why? It's not obvious to me that such an obscure interaction between two unrelated extensions is better than nothing.

Ambrevar:
The title is not an extension, it is a parameter of the HTML renderer. Basically, this would lead to more automation from the generator: start a page with

```
% My title
```

And the `<title>My title</title> will be set without any additional code needed on the user side.

rtfb:
But the `HTMLRendererParameters.Title` is a part of an output only when using `CompletePage` extension. And then you consider a title only when a `Titleblock` extension is on. That couples these extensions in a very non-straightforward way.

Ambrevar:
Right, forgot that I had completepage on. OK, forget this PR.

PR #298, opened 2016-08-10
v2: Avoid useless AST splits with AutoLink

Ambrevar:


inline.go
```diff
@@ -75,6 +76,9 @@ func (p *parser) inline(data []byte) {
 			if p.inlineCallback[data[end]] != nil {
 				if end+1 < len(data) && data[end] == 'h' && data[end+1] != 't' {
 					end++
+				} else if strings.ContainsRune("hHmMfF", rune(data[end])) && !isAutoLink(p, data, end) {
```
rtfb: Is there a reason to have `ContainsRune` and not `IndexByte` here?

rtfb:
Was this the only source of text splits?

inline.go
Ambrevar: ContainsRune returns a boolean, IndexByte returns the position. While you can achieve the same result with both, I believe it the be more idiomatic to use the function that bears the name of what you want to do,

Ambrevar:
Excellent question. Now I am looking at the callbacks, I just noticed that useless splits can happen pretty much all the time:
- maybeImage
- emphasis
- entity
- ...
More generally, any time a callback can potentially return 0.

Don't merge this PR, we need to find a more generic solution.

inline.go
rtfb: I would agree this is more readable if not for the cast of byte to rune, which spoils all the gain, if there was any.

But the style is a secondary question to me here, performance is the primary. This is one of the hottest paths in the library and we're calling a (potentially) more expensive call for nearly every byte of input. Of course, `ContainsRune` is implemented in terms of `IndexByte`, so there should not be much difference asymptotically, but the little constant overhead might be very noticeable.

Ambrevar: Agreed. The compiler can inline the function call however, and the cast does not cost anything as far as I know. 

Anyways, I'm working on a generic fix for the useless splits issue which won't require this line :)

Ambrevar:
See #301.

PR #300, opened 2016-08-12
v2: Remove the LaTeX renderer stub and point to Ambrevar's implementation

Ambrevar:
I've implemented an almost fully functional [LaTeX renderer](https://bitbucket.org/ambrevar/blackfriday-latex).

The main extension that is broken at the moment is footnote support, but this needs upatream changes to the parser first.


rtfb:
What kind of changes are required in the parser? Did it work in v1?

Ambrevar:
Footnotes suffer from the same issue as for the TOC: the HTML-specific code needs to be moved to the HTML renderer.

markdown.go
```diff
@@ -15,7 +15,8 @@
 
 // Package blackfriday is a markdown processor.
 //
-// Translates plain text with simple formatting rules into HTML or LaTeX.
+// Translates plain text with simple formatting rules into an intermediate
+// representation that can be renderer to HTML.
```
shurcooL: renderer -> renderered.

Maybe also say "to HTML or other formats."

markdown.go
```diff
@@ -160,7 +161,7 @@ var blockTags = map[string]struct{}{
 // If the callback returns false, the rendering function should reset the
 // output buffer as though it had never been called.
 //
-// Currently HTML and Latex implementations are provided
+// Only HTML implementation is provided.
```
shurcooL:
```
// Only HTML implementation is provided in this repository,
// see readme for external implementations.
```

shurcooL:
I am personally really really glad to see this change! It makes the core repo smaller and more focused on Markdown itself. 🎉 

LGTM.

markdown.go
Ambrevar: renderered -> rendererered :)
I'll fix the typos just now.

Ambrevar:
Please have a look at the new LaTeX renderer and comments: I believe it to be a nice benchmark for the current development made on v2, e.g. to see how flexible the new interface is, or what is left to improve.

rtfb:
> Footnotes suffer from the same issue as for the TOC: the HTML-specific code needs to be moved to the HTML renderer.

Are you going to do that yourself? Maybe it would make sense to land this PR after footnotes are fixed.

Ambrevar:
Don't know if I'll time for this. I guess whoever wrote the footnote code would be the best fit for fixing this quickly.

rtfb:
OK, so I've given it a brief look and I have mixed news about it.

First, footnotes are also stubbed out in v1 Latex renderer. So we're OK to merge this PR functionality-wise, our only concern is API completeness.

Now from the API standpoint parsing references and footnotes are very tightly coupled with the parser at the moment. Moving that to the renderer would be a big effort. Here are some highlights of that effort:

* Footnote/reference extraction is happening in `firstPass`, which we would like to eliminate altogether.
* I want AST to be a single structure to be passed from the parser to the renderer, so the references have to be reflected in the AST somehow. Currently, they have two(!) separate data structures embedded inside the parser.

It seems to me that the cleanest way to implement the refs and footnotes would be to move their detection from `firstPass` to inline parsing stage and to hold them as a part of `Link` nodes in the tree.

Ambrevar:
I was expecting some complex intricacis like this, which is why I was a bit reluctant to dive into this issue! :/

I fully agreed with your current conclusions.

Question: should we reuse the 'Link' type to store footnotes or use a dedicated type?

rtfb:
Take a look at #302. It turned out to be somewhat easier than I anticipated.

rtfb:
Looks like the move of package level docs has broken this patch. Please fix it when you can and I'll merge.

Ambrevar:
Wait, I solved the conflict the wrong way...

Ambrevar:
Should be good to go now.

rtfb:
Thanks, @Ambrevar!

PR #301, opened 2016-08-18
v2: Only split when inline callbacks consume some bytes

Ambrevar:
The former hacks around maybeLineBreak and Smartypants are no longer
needed.
The algorithm has been streamlined: shorter, simpler, faster.
The 'currBlock' field of the parser is gone.

Bonus: A rough perf increase of 6-8%!

### DIFF
--- a/inline.go
+++ b/inline.go
@@ -192,10 +192,8 @@ func maybeLineBreak(p *parser, data []byte, offset int) (int, *Node) {
 // newline without two spaces works when HardLineBreak is enabled
 func lineBreak(p *parser, data []byte, offset int) (int, *Node) {
 	if p.flags&HardLineBreak != 0 {
-		// log.Print("LINEBREAK")
 		return 1, NewNode(Hardbreak)
 	}
-	// log.Print("NOLINEBREAK")
 	return 0, nil
 }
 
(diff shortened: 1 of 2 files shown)


inline.go
```diff
@@ -189,35 +167,36 @@ func codeSpan(p *parser, data []byte, offset int) int {
...
+	return 0, nil
 }
 
 // newline without two spaces works when HardLineBreak is enabled
-func lineBreak(p *parser, data []byte, offset int) int {
+func lineBreak(p *parser, data []byte, offset int) (int, *Node) {
 	if p.flags&HardLineBreak != 0 {
-		p.currBlock.AppendChild(NewNode(Hardbreak))
-		return 1
+		// log.Print("LINEBREAK")
```
rtfb: nit: debug log

inline.go
```diff
@@ -189,35 +167,36 @@ func codeSpan(p *parser, data []byte, offset int) int {
...
+		return 1, NewNode(Hardbreak)
 	}
-	return 0
+	// log.Print("NOLINEBREAK")
```
rtfb: one more

rtfb:
This is brilliant! Thanks a lot!

Please fix the couple stray debug logs and let's merge it.

inline.go
Ambrevar: That's odd, I recall removing those 2 lines...

Ambrevar:
Done.
Please squash-merge.

rtfb:
Thanks again! :-)

PR #302, opened 2016-09-03
v2: move footnotes to html

rtfb:
Clean up footnotes part of an AST: don't force HTML-specific pieces there, just keep a clean list of footnotes. Since some renderers might want to process footnotes differently, let them know about footnotes by having a flag on that list.

### DIFF
--- a/markdown.go
+++ b/markdown.go
@@ -416,9 +416,8 @@ func (p *parser) parseRefsToAST() {
 		return
 	}
 	p.tip = p.doc
-	finalizeHTMLBlock(p.addBlock(HTMLBlock, []byte(`<div class="footnotes">`)))
-	p.addBlock(HorizontalRule, nil)
 	block := p.addBlock(List, nil)
+	block.IsFootnotesList = true
 	block.ListFlags = ListTypeOrdered
 	flags := ListItemBeginningOfList
 	// Note: this loop is intentionally explicit, not range-form. This is
@@ -441,7 +440,6 @@ func (p *parser) parseRefsToAST() {
 	above := block.Parent
 	finalizeList(block)
 	p.tip = above
-	finalizeHTMLBlock(p.addBlock(HTMLBlock, []byte("</div>")))
 	block.Walk(func(node *Node, entering bool) WalkStatus {
 		if node.Type == Paragraph || node.Type == Header {
 			p.inline(node, node.content)
--- a/markdown.go
+++ b/markdown.go
@@ -453,7 +453,6 @@ func (p *parser) parseRefsToAST() {
 
 // first pass:
 // - normalize newlines
-// - extract references (outside of fenced code blocks)
 // - expand tabs (outside of fenced code blocks)
 // - copy everything else
 func firstPass(p *parser, input []byte) []byte {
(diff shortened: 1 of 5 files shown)


markdown.go
```diff
@@ -565,13 +535,50 @@ func secondPass(p *parser, input []byte) {
...
+//     <p><a href=\"/url/\" title=\"title\">link</a></p>
+//
+// And a reference structure will be populated as follows:
+//
+//     p.refs["ref"] = &reference{
+//         link: "/url/",
+//         title: "tooltip title",
+//     }
+//
+// Alternatively, reference can contain an information about a footnote.
```
Ambrevar: 'an information' => 'information'

markdown.go
```diff
@@ -394,8 +394,7 @@ func Parse(input []byte, opts Options) *Node {
 		p.notes = make([]*reference, 0)
 	}
 
-	first := firstPass(p, input)
-	secondPass(p, first)
+	p.block(firstPass(p, input))
```
Ambrevar: If secondPass is gone, firstPass becomes a confusing name.

Ambrevar:
I had a very short glimpse at this: this looks great! The elimination of the secondPass is big plus.
I'll have a closer look later and implement the LaTeX rendering.

markdown.go
rtfb: Fixed.

markdown.go
rtfb: Renamed to `preprocess`.

rtfb:
Hey @Ambrevar, did you have a chance to give it a longer glimpse? Is it good to land?

Ambrevar:
No, I haven't, I am very busy at the moment. Can't promise anything. If you feel like merging, go for it. If something goes wrong in terms of interfacing with the LaTeX renderer, we can still fix it afterwards.

rtfb:
No worries, I'll go ahead and merge. Thanks!

Ambrevar:
I've implemented footnotes on the LaTeX side. It works. There is room for discussion though.

In LaTeX, a footnote is rendered as `\footnote{content...}`. As simple as that. There are other possibilities such as `\footnotetext` and `\footnotemark` but those are excluded since they cannot guarantee that the mark and the text will be on the same page.

The footnote text is currently stored at the end of the AST. A first-pass is needed in order to make the link between the footnote mark and the footnote text (e.g. with a map) so that the complete footnote can be rendered when the renderer walks through a footnote mark.

My suggestion: shouldn't we move the footnote text nodes to a dedicated `parser.Footnotes map[string]*Node`? This would make footnote rendering more straightforward for all renderers and thus more universal.

rtfb:
> My suggestion: shouldn't we move the footnote text nodes to a dedicated parser.Footnotes map[string]*Node? This would make footnote rendering more straightforward for all renderers and thus more universal.

I don't like this proposal for two reasons: first, having a single output from the parse stage, an AST, seems cleaner. Second, `parser` struct is currently not exposed and the prospect of exposing it doesn't appeal either.

Instead, I was thinking of leaving the footnote tree nodes where they are, but having a pointer to them from the referring `Node`, in the `LinkData` payload. Does that sound good?

Ambrevar:
I guess that would do :)

PR #303, opened 2016-09-04
v2: Extract package level documentation to doc.go

rtfb:
Plus elaborate the documentation a bit.


doc.go
```diff
@@ -0,0 +1,18 @@
+// Package blackfriday is a markdown processor.
+//
+// Translates plain text with simple formatting rules into an AST, which can
```
shurcooL: Maybe "It translates ..."?

shurcooL:
See one comment, otherwise LGTM.

PR #304, opened 2016-09-04
v2: Run the reference benchmark on Travis

rtfb:
So that we could have a historic record of performance.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,4 @@ install:
 
 script:
     - go test -v ./...
+    - go test -short -bench BenchmarkReference -benchmem
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,4 +16,4 @@ install:
 
 script:
     - go test -v ./...
-    - go test -short -bench BenchmarkReference -benchmem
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -16,3 +16,4 @@ install:
 
 script:
     - go test -v ./...
+    - go test -short -bench BenchmarkReference -benchmem
```
shurcooL: What is `-short` flag for?

You probably want this line to only run that `BenchmarkReference` benchmark and no tests, right?

If so, I think you should add `-run=^$` or something equivalent.

rtfb: You're right, fixed.

.travis.yml
```diff
@@ -16,3 +16,4 @@ install:
 
 script:
     - go test -v ./...
+    - go test -run=^$ -bench BenchmarkReference -benchmem
```
shurcooL: Really minor and optional, but I'd consider doing including `=` in `-bench=BenchmarkReference` for consistency with `-run=^$`. Up to you.

shurcooL:
Minor optional comment, LGTM.

PR #306, opened 2016-09-17
V2 add links to footnotes

rtfb:
As per discussion in #302, this adds a pointer to footnote nodes from the referring Node, in the LinkData payload.

### DIFF
--- a/markdown.go
+++ b/markdown.go
@@ -213,14 +213,16 @@ func (p *parser) finalize(block *Node) {
 }
 
 func (p *parser) addChild(node NodeType, offset uint32) *Node {
-	for !p.tip.canContain(node) {
+	return p.addExistingChild(NewNode(node), offset)
+}
+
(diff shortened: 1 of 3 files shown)


rtfb:
Bump. Anyone to eyeball this?

markdown.go
```diff
@@ -213,14 +213,16 @@ func (p *parser) finalize(block *Node) {
 }
 
 func (p *parser) addChild(node NodeType, offset uint32) *Node {
-	for !p.tip.canContain(node) {
+	return p.addExistingChild(NewNode(node), offset)
```
shurcooL: There are only 3 callers of `addChild` now, and it differs from `addExistingChild` only by `NewNode`.

See if it's better to keep it or remove it. Up to you, either is fine. Just wanted to raise your attention here.

rtfb: I'll keep an eye on where it goes, thanks!

PR #310, opened 2016-10-05
v2: Fix walk with non-container root nodes

Ambrevar:
When passed a non-container root node, the former algorithm would go on
walking down the rest of the tree beyond the root.

The former walk fix was supposed to do that but somehow the code
disappeared in the process.

### DIFF
--- a/node.go
+++ b/node.go
@@ -309,7 +309,7 @@ func newNodeWalker(root *Node) *nodeWalker {
 }
 
 func (nw *nodeWalker) next() {
-	if !nw.entering && nw.current == nw.root {
+	if (!nw.current.isContainer() || !nw.entering) && nw.current == nw.root {
 		nw.current = nil
 		return
 	}


rtfb:
> When passed a non-container root node, the former algorithm would go on
walking down the rest of the tree beyond the root.

I'm a bit confused about what's the problem case here. A non-container node is a leaf node, it should have no children, so there should be nothing to traverse below it. If you encountered a case where non-container node has children, let's fix that, not the traverser.

Ambrevar:
Err... no, that's not what I meant. Concrete example

```
- document
-- text
-- list
--- paragraph
---- text
```

In the above tree, walking the first 'text' leaf-node should walk that node only, right? The current algorithm will step out and walk down the list, paragraph, etc. until the end of the tree.

rtfb:
Ah, got it. Yes, that was a bug. In it goes.

PR #322, opened 2016-11-24
V2 perf tweaks

rtfb:


rtfb:
Anybody interested in giving it another round of eyeballing? If we're good, I'll go ahead and merge.

block.go
```diff
@@ -1072,32 +1086,37 @@ func (p *parser) oliPrefix(data []byte) int {
...
 }
 
 // returns definition list item prefix
 func (p *parser) dliPrefix(data []byte) int {
+	if len(data) < 2 {
+		return 0
+	}
 	i := 0
-
 	// need a : followed by a spaces
```
shurcooL: Update comment to mention tabs.

block.go
```diff
@@ -1175,6 +1194,9 @@ func finalizeList(block *Node) {
 func (p *parser) listItem(data []byte, flags *ListType) int {
 	// keep track of the indentation of the first line
 	itemIndent := 0
+	if data[itemIndent] == '\t' {
```
shurcooL: Can be simplified to `if data[0] == '\t' {`.

block.go
```diff
@@ -1243,11 +1265,18 @@ gatherlines:
...
+		indentIndex := 0
+		if data[line] == '\t' {
+			indentIndex++
+			indent += 4
+		} else {
+			for indent < 4 && line+indent < i && data[line+indent] == ' ' {
+				indent++
+				indentIndex++
+			}
 		}
```
shurcooL: Is it neccessary to put `for` loop inside `else` block, or would the behavior be the same if `for` block would simply follow the `if` statement?

The reason I ask is because it looks a little inconsistent with line 1178/1200 (beginning of `parser.listItem` method), and I wonder if we should make it consistent (either by moving that `for` into `else`, or by moving this `for` out of the `else`).

block.go
```diff
@@ -1404,7 +1436,8 @@ func (p *parser) paragraph(data []byte) int {
 			// did this blank line followed by a definition list item?
 			if p.flags&DefinitionLists != 0 {
 				if i < len(data)-1 && data[i+1] == ':' {
-					return p.list(data[prev:], ListTypeDefinition)
+					ret := p.list(data[prev:], ListTypeDefinition)
+					return ret
```
shurcooL: Why this change? What does `ret` mean?

esc_test.go
```diff
@@ -0,0 +1,48 @@
...
+		[]byte("4 < 5."),
+		[]byte("6 > 5."),
+		[]byte("Here's a [link] [1] with an ampersand in the URL."),
+		[]byte("Here's a link with an ampersand in the link text: [AT&T] [2]."),
+		[]byte("Here's an inline [link](/script?foo=1&bar=2)."),
+		[]byte("Here's an inline [link](</script?foo=1&bar=2>)."),
+		[]byte("[1]: http://example.com/?foo=1&bar=2"),
+		[]byte("[2]: http://att.com/  \"AT&T\""),
+	}
+	var buff bytes.Buffer
```
shurcooL: This variable is mostly commonly called `buf`, not `buff`.

https://github.com/golang/go/search?q=buf - 652 results
https://github.com/golang/go/search?q=buff - 3 results

html.go
```diff
@@ -457,30 +552,36 @@ func (r *HTMLRenderer) RenderNode(w io.Writer, node *Node, entering bool) WalkSt
...
-				r.out(w, tag("/tt", nil, false))
+				r.out(w, ttCloseTag)
 			}
 		} else {
 			if entering {
 				dest = r.addAbsPrefix(dest)
-				//if (!(options.safe && potentiallyUnsafe(node.destination))) {
-				attrs = append(attrs, fmt.Sprintf("href=%q", esc(dest)))
-				//}
+				var hrefBuff bytes.Buffer
```
shurcooL: Same here, I'd name it `hrefBuf`.

inline.go
```diff
@@ -728,25 +715,38 @@ func linkEndsWithEntity(data []byte, linkEnd int) bool {
 	return entityRanges != nil && entityRanges[len(entityRanges)-1][1] == linkEnd
 }
 
+var prefixes = [][]byte{
```
shurcooL: This variable name was okay when it was inside `maybeAutoLink` func, but it's not okay as a package level variable. It's not descriptive at all.

Call it `autoLinkPrefixes` maybe.

inline.go
```diff
@@ -728,25 +715,38 @@ func linkEndsWithEntity(data []byte, linkEnd int) bool {
...
+	[]byte("http://"),
+	[]byte("https://"),
+	[]byte("ftp://"),
+	[]byte("file://"),
+	[]byte("mailto:"),
+}
+
+func hasPrefixCaseInsensitive(s, prefix []byte) bool {
```
shurcooL: Mind documenting this new helper?

```
// hasPrefixCaseInsensitive reports <what it reports>
```

markdown.go
```diff
@@ -442,63 +442,6 @@ func (p *parser) parseRefsToAST() {
 	})
 }
 
-// preprocess does a preparatory first pass over the input:
-// - normalize newlines
-// - expand tabs (outside of fenced code blocks)
-// - copy everything else
-func preprocess(p *parser, input []byte) []byte {
```
shurcooL: Whooo! 🔥  🎉

block.go
```diff
@@ -769,7 +764,9 @@ func (p *parser) table(data []byte) int {
 		}
 
 		// include the newline in data sent to tableRow
-		i++
+		if i < len(data)-1 && data[i] == '\n' {
```
rtfb: This must've been a copypasta, the tests still pass with `-1` removed. Thanks for catching :-)

block.go
```diff
@@ -935,11 +941,11 @@ func (p *parser) tableRow(data []byte, columns []CellAlignFlags, header bool) {
...
 func (p *parser) quotePrefix(data []byte) int {
 	i := 0
-	for i < 3 && data[i] == ' ' {
+	for i < 3 && i < len(data) && data[i] == ' ' {
 		i++
 	}
-	if data[i] == '>' {
-		if data[i+1] == ' ' {
+	if i < len(data) && data[i] == '>' {
+		if i < len(data)-1 && data[i+1] == ' ' {
```
rtfb: Yup

block.go
rtfb: Ohh, that's an oversight. I must've done it to debug-print and forgot to change back.

block.go
rtfb: The spec says "List markers must be followed by one or more spaces *or* a tab" (emphasis mine), so I unified into the `else` clause in both places. Thanks!

inline.go
rtfb: I went with `protocolPrefixes`.

PR #324, opened 2016-11-24
Move TOC and OmitContents to HTML flags

### LINKED ISSUE #277: v2: Fix extensions passing to renderer and main Markdown func
It is possible to pass contradictory extensions to the renderer and the main `Markdown` function. And even if they're not contradictory, the duplication is confusing:

```
 renderer := bf.NewHTMLRenderer(bf.HTMLRendererParameters{
         Extensions: bf.NoExtensions,
 })
 withmath := bf.Markdown(md, renderer, bf.Options{Extensions: bf.CommonExtensions})
```

rtfb:
The root problem this commit fixes is the duplication of Extensions
field in HTMLRendererParameters. The duplication crept in there only to
support these two flags, so moving the flags solves the problem. They're
only used in renderer anyway.

Fixes #277.

### DIFF
--- a/block_test.go
+++ b/block_test.go
@@ -1601,7 +1601,9 @@ func TestTOC(t *testing.T) {
 		"#",
 		"",
 	}
-	doTestsBlock(t, tests, TOC)
+	doTestsParam(t, tests, TestParams{
+		HTMLFlags: UseXHTML | TOC,
+	})
 }
 
 func TestOmitContents(t *testing.T) {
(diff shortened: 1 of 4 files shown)


block_test.go
```diff
@@ -1601,7 +1601,9 @@ func TestTOC(t *testing.T) {
 		"#",
 		"",
 	}
-	doTestsBlock(t, tests, TOC)
+	doTestsParam(t, tests, TestParams{
+		HTMLFlags: UseXHTML | TOC,
```
shurcooL: Is the change to set `UseXHTML` unrelated to the fix?

markdown.go
```diff
@@ -46,8 +46,6 @@ const (
 	AutoHeaderIDs                                 // Create the header ID from the text
 	BackslashLineBreak                            // Translate trailing backslashes into line breaks
 	DefinitionLists                               // Render definition lists
-	TOC                                           // Generate a table of contents
-	OmitContents                                  // Skip the main contents (for a standalone table of contents)
```
shurcooL: Yeah, these extensions have nothing to do with parsing Markdown, so it makes sense to move them from here. 👍

block_test.go
rtfb: It's a bit obscure: `doTestsBlock` passes `UseXHTML` down to `doTestsParam` and since I'm switching to this "lower level" func, the flag bubbled up.

shurcooL: Gotcha, thanks. I didn't notice the change from `doTestsBlock` to `doTestsParam`.

PR #327, opened 2017-01-26
Update README for v2

rtfb:
Properly document v2. This should also get ported to master.

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,9 +273,9 @@ License
 [Blackfriday is distributed under the Simplified BSD License](LICENSE.txt)
 
 
-   [1]: http://daringfireball.net/projects/markdown/ "Markdown"
-   [2]: http://golang.org/ "Go Language"
+   [1]: https://daringfireball.net/projects/markdown/ "Markdown"
+   [2]: https://golang.org/ "Go Language"
    [3]: https://github.com/vmg/sundown "Sundown"
    [4]: https://godoc.org/gopkg.in/russross/blackfriday.v2#Parse "Parse func"
    [5]: https://github.com/microcosm-cc/bluemonday "Bluemonday"
-   [6]: http://labix.org/gopkg.in "gopkg.in"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,23 +16,44 @@ It started as a translation from C of [Sundown][3].
...
+* Cleaned up API
+* A separate call to [`Parse`][4], which produces an abstract syntax tree for
+  the document
+* Latest bug fixes
+* Flexibility to easily add your own rendering extensions
+
+Potential drawbacks:
+
+* Our benchmarks show v2 to be slightly slower than v1. Currently in the
+  ballpark of around 15%.
```
shurcooL: Is this before or after #322? With many of the optimizations such as removing the preprocess step, using `io.Writer` and predeclared `[]byte` slices, what's causing v2 to still be slower?

README.md
```diff
@@ -256,3 +276,6 @@ License
    [1]: http://daringfireball.net/projects/markdown/ "Markdown"
    [2]: http://golang.org/ "Go Language"
    [3]: https://github.com/vmg/sundown "Sundown"
+   [4]: https://godoc.org/gopkg.in/russross/blackfriday.v2#Parse "Parse func"
+   [5]: https://github.com/microcosm-cc/bluemonday "Bluemonday"
+   [6]: http://labix.org/gopkg.in "gopkg.in"
```
shurcooL: > http://labix.org/gopkg.in "gopkg.in"

Use `https://` please.

README.md
```diff
@@ -256,3 +276,6 @@ License
    [1]: http://daringfireball.net/projects/markdown/ "Markdown"
    [2]: http://golang.org/ "Go Language"
```
shurcooL: Might as well update these 2 old links to use https too. They both support it.

README.md
rtfb: After. Before that, it was around 2.5x slower. The remaining slowdown is very washed out, couldn't pick it out yet.

PR #328, opened 2017-02-02
Change the public interface to use functional options

rtfb:
Convert the most important Blackfriday's function, `Markdown()`, to accept functional options (as per this Dave Cheney's post: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis)

This dawned on me a bit late in the cycle, but thankfully just in time before we announce v2 with full lungs.

In essence, here's the main benefit: instead of calling `blackfriday.MarkdownCommon(input)`, you can now call `blackfriday.Markdown(input)`.

Instead of `blackfriday.MarkdownBasic(input)`, call `blackfriday.Markdown(input, WithNoExtensions())`.

And for a fully customized version:

``` go
blackfriday.Markdown(input, blackfriday.WithRefOverride(override), blackfriday.WithExtensions(exts), blackfriday.WithRenderer(yourRenderer))
```

I am also very tempted to take this chance to change the `input []byte` to `io.Reader`. Unfortunately, under the hood that would take a lot of changes and I can't do them quickly. What do you think, would it make sense to make the interface change now, while we're breaking compatibility anyway, but leave the implementation buffered as it is now?

### DIFF
--- a/block.go
+++ b/block.go
@@ -34,7 +34,7 @@ var (
 // Parse block-level data.
 // Note: this function and many that it calls assume that
 // the input buffer ends with a newline.
-func (p *parser) block(data []byte) {
+func (p *Parser) block(data []byte) {
 	// this is called recursively: enforce a maximum depth
 	if p.nesting >= p.maxNesting {
 		return
@@ -197,14 +197,14 @@ func (p *parser) block(data []byte) {
 	p.nesting--
 }
 
-func (p *parser) addBlock(typ NodeType, content []byte) *Node {
+func (p *Parser) addBlock(typ NodeType, content []byte) *Node {
 	p.closeUnmatchedBlocks()
 	container := p.addChild(typ, 0)
 	container.content = content
 	return container
 }
 
-func (p *parser) isPrefixHeader(data []byte) bool {
+func (p *Parser) isPrefixHeader(data []byte) bool {
 	if data[0] != '#' {
 		return false
 	}
@@ -221,7 +221,7 @@ func (p *parser) isPrefixHeader(data []byte) bool {
 	return true
 }
 
-func (p *parser) prefixHeader(data []byte) int {
+func (p *Parser) prefixHeader(data []byte) int {
 	level := 0
 	for level < 6 && level < len(data) && data[level] == '#' {
 		level++
@@ -267,7 +267,7 @@ func (p *parser) prefixHeader(data []byte) int {
 	return skip
 }
 
-func (p *parser) isUnderlinedHeader(data []byte) int {
+func (p *Parser) isUnderlinedHeader(data []byte) int {
 	// test of level 1 header
 	if data[0] == '=' {
 		i := skipChar(data, 1, '=')
@@ -291,7 +291,7 @@ func (p *parser) isUnderlinedHeader(data []byte) int {
 	return 0
 }
 
-func (p *parser) titleBlock(data []byte, doRender bool) int {
+func (p *Parser) titleBlock(data []byte, doRender bool) int {
 	if data[0] != '%' {
 		return 0
 	}
@@ -315,7 +315,7 @@ func (p *parser) titleBlock(data []byte, doRender bool) int {
 	return consumed
 }
 
-func (p *parser) html(data []byte, doRender bool) int {
+func (p *Parser) html(data []byte, doRender bool) int {
 	var i, j int
 
 	// identify the opening tag
@@ -419,7 +419,7 @@ func finalizeHTMLBlock(block *Node) {
 }
 
 // HTML comment, lax form
-func (p *parser) htmlComment(data []byte, doRender bool) int {
+func (p *Parser) htmlComment(data []byte, doRender bool) int {
 	i := p.inlineHTMLComment(data)
 	// needs to end with a blank line
 	if j := p.isEmpty(data[i:]); j > 0 {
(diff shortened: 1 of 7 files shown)


shurcooL:
I would probably rewrite that customized version to not scroll off-screen:

```Go
blackfriday.Markdown(input,
	blackfriday.WithRefOverride(override),
	blackfriday.WithExtensions(exts), blackfriday.WithRenderer(yourRenderer))
```

Just to give me a better idea of what it'd look like.

I have to admit this style hasn't fully grown on me yet, but I don't mind it. SGTM.

One thing, see an excellent writeup by @rakyll at https://rakyll.org/style-packages/, the "Provide examples to fill the gaps" section (btw @rakyll, I wish I could link to individual sections).

I think we should definitely provide examples in godoc that display all the common use cases with the new functional options approach. Both for helping people understand what's possible, and for convenience of copy pasting a good starting point.

From what I can tell, you've already done so in the documentation for `Markdown` func, which is 💯! I think it'd be a good idea to do it in the package comment as well. That way it'll show up on top of godoc where the package is introduced and described.

---

I am also very tempted to take this chance to change the input `[]byte` to `io.Reader`.

Providing an `io.Reader` is always easier than `[]byte`, right? If you have `[]byte`, you can wrap with `bytes.NewReader`, which is inexpensive (although copying out of that reader into `[]byte` is a little wasteful). If you have `io.Reader`, you pass it directly in. Providing `[]byte` when you have `io.Reader` requires reading it all first, which would be wasteful if the library doesn't need all bytes at once to make progress.

Can you remind me if we have plans to eventually do some streaming and on-the-fly processing, rather than simply dumping the entire `io.Reader` content into `[]byte` at the very beginning?

If so, then making that change of API to take `io.Reader` now, dump it to `[]byte` as first internally _for now_, and potentially doing more streaming later also SGTM.

Would it be a good or bad idea to provide two funcs in the API, one that takes `[]byte` and the other `io.Reader`?

rakyll:
(@shurcooL, there are anchor links in the source but the titles are not made clickable by the theme. I am going to fix it today. See https://rakyll.org/style-packages/#provide-examples-to-fill-the-gaps. Thanks!)

shurcooL:
About functional options, have you also considered just an `Options` struct? I.e.:

```Go
func Markdown(input []byte, opt *Options) { ... }

type Options struct {
	RefOverride ...
	Extensions ...
	Renderer ...
}
```

I guess that's similar to what we have with `MarkdownOptions` right now:

https://godoc.org/github.com/russross/blackfriday#MarkdownOptions

Except we could move `Renderer` parameter into options too (and possibly give it a sane default, making it optional).

rtfb:
> About functional options, have you also considered just an `Options` struct?

Yes. Dave's article has a detailed rationale against that, but in short: option-less call becomes cumbersome. One would have to provide a `nil` or an `Options{}` while with functional options there's literally nothing.

---

WRT the `io.Reader`, I realized there's no point. The real win of having a streaming interface is when we can connect an `io.Reader` on one end with an `io.Writer` on the other without any buffering in between. But in our case this won't happen, because we produce an AST in the middle. And while a typical AST is quite shallow, list-like, things like TOC and refrence-style links make it necessary to buffer an entire document.

So, no streaming for Blackfriday :-/

shurcooL:
> Yes.

Fair enough.

> Dave's article has a detailed rationale against that, but in short: option-less call becomes cumbersome. One would have to provide a nil or an Options{} while with functional options there's literally nothing.

Is that really the main reason? To me, seeing this in godoc:

```Go
func Process(..., opt *Options) ([]byte, error)
```

Is very very clear and readable, because it's so explicit. I can immediately look at the `Options` struct and instantly understand all of the options available:

```Go
type Options struct {
    Fragment  bool // Accept fragment of a source file (no package statement)
    AllErrors bool // Report all errors (not just the first 10 on different lines)

    Comments  bool // Print comments (true if nil *Options provided)
    TabIndent bool // Use tabs for indent (true if nil *Options provided)
    TabWidth  int  // Tab width (8 if nil *Options provided)

    FormatOnly bool // Disable the insertion and deletion of imports
}
```

In the end, for me, I find an `Options` struct much easier to write, and very easy to read. Functional options are somewhat easy to read (I have to guess that magic behind the scenes does the right thing, but I can't be as certain of it without spending more time looking at its internals), and more hard to write (which is why I think documenting some examples is so important).

However, that's just my POV. I've used options structs a lot, and functional options almost never. They might grow on me after more exposure, I don't know.

I'm not against trying them here, I just wanted to better understand the tradeoffs.

---

> The real win of having a streaming interface is when we can connect an io.Reader on one end with an io.Writer on the other without any buffering in between.

Precisely.

> But in our case this won't happen, because we produce an AST in the middle. And while a typical AST is quite shallow, list-like, things like TOC and refrence-style links make it necessary to buffer an entire document.

Makes sense. I suspected that could be the case. The ability to stream comes with hard requirement that limit what you're able to do in terms of processing, and Markdown (with extensions) is too powerful of a language to be able to get away without needing buffering.

Ambrevar:
I like functional options.

What about #277 then?

Streaming: what about streaming possibilities if options like refs/TOC are off?

rtfb:
> Streaming: what about streaming possibilities if options like refs/TOC are off?

Reference style links are standard, and we can't know whether an input document contains them until we parse it all, so, however disappointing that might sound, streaming is not an option for markdown processing... :-(

Ambrevar:
Right, forgot about that.

v2 inbound, then.

shurcooL:
Perhaps relevant, this discourse happened recently. https://twitter.com/rogpeppe/status/890483872605057024.

Specifically, the tweet refers to package [`cmp`](https://godoc.org/github.com/google/go-cmp/cmp).

rtfb:
A very neat use of functional options indeed! And seems to be a useful package in general, thanks.

Issue #311, opened 2016-10-05
Discussion: v2 performance tweaks


rtfb:
Hi all,

I have recently pushed a branch, [v2-perf-tweaks](https://github.com/russross/blackfriday/tree/v2-perf-tweaks), with some initial performance improvements for v2. I'm opening this discussion thread for discussing whatever issues people might find reviewing that code and/or suggestions/observations for further improvement.

rtfb:
As a seed for discussion I want to bring your attention to 993325d. It gets rid of the general purpose `html.EscapeString`-ing and rolls our own version. This gives a significant performance boost (~25% faster on CPU, ~45% less allocs), but is significantly narrower in scope. It only deals with `&`, `"`, `<` and `>`, like v1 did, and will leave anything else alone.

I want other people's opinions on whether this is a reasonable trade-off or not.

shurcooL:
> is significantly narrower in scope. It only deals with `&`, `"`, `<` and `>`, like v1 did, and will leave anything else alone.

In what way is the scope of `html.EscapeCompare` larger? What does it do that `escapeHTML` doesn't? I only see a difference in handling of the `'` character.

rtfb:
Hm. Looking at its code, you're right, the only difference is that it also handles the `'`. I must have mixed up `html.EscapeString` and `html.UnescapeString` in my head -- the latter is *much* heavier beast. Well, then this change is less controversial than I thought.

Give all of them some good scrutiny, though :-)

Ambrevar:
Nice job! Runs twice as fast for me!

 Try out the suggestions in my code review: they might save some more cycles.

Ambrevar:
Note that you can use the same table trick for `html.go/cellAlignment`:

``` go
var cellAlignment = [4]byte{
	0: 'l',
	bf.TableAlignmentLeft:   'l',
	bf.TableAlignmentRight:  'r',
	bf.TableAlignmentCenter: 'c',
}
```

This will not save as much though since it is just a switch.

rtfb:
> Note that you can use the same table trick for `html.go/cellAlignment`:

I'll see to it, but it's very far from a critical path, so I don't expect it to have any performance influence and will pick the one that's more readable.

rtfb:
120bb2f gets rid of the preprocess stage. Yay!

[...]

Issue #173, opened 2015-06-03
index out of range panic in isReference


opennota:
The below program panics:
```
package main

import "github.com/russross/blackfriday"

func main() {
	renderer := blackfriday.HtmlRenderer(blackfriday.HTML_USE_XHTML, "", "")
	blackfriday.Markdown([]byte("   ["), renderer,
		blackfriday.EXTENSION_TABLES|
			blackfriday.EXTENSION_AUTOLINK|
			blackfriday.EXTENSION_FOOTNOTES|
			blackfriday.EXTENSION_TITLEBLOCK|
			blackfriday.EXTENSION_HEADER_IDS|
			blackfriday.EXTENSION_FENCED_CODE|
			blackfriday.EXTENSION_SPACE_HEADERS|
			blackfriday.EXTENSION_STRIKETHROUGH|
			blackfriday.EXTENSION_TAB_SIZE_EIGHT|
			blackfriday.EXTENSION_AUTO_HEADER_IDS|
			blackfriday.EXTENSION_LAX_HTML_BLOCKS|
			blackfriday.EXTENSION_HARD_LINE_BREAK|
			blackfriday.EXTENSION_NO_INTRA_EMPHASIS|
			blackfriday.EXTENSION_BACKSLASH_LINE_BREAK|
			blackfriday.EXTENSION_NO_EMPTY_LINE_BEFORE_BLOCK,
	)
}
```

Issue #172, opened 2015-06-03
index out of range panic in scanLinkRef


opennota:
The below program panics:
```
package main

import "github.com/russross/blackfriday"

func main() {
	blackfriday.MarkdownCommon([]byte("[]:<"))
}
```

PR #331, opened 2017-02-12
Fix #180 again

### LINKED ISSUE #180: Newline required after last "reference style link" or it is not parsed.

rtfb:
The check was introduced with d28de22, when fixing #172 and #173.
Then I removed it with bcd6dd8 when fixing #180
And then it was reintroduced with 232d06c when fixing regression.

It seems that the check can be removed again. All these cases now have
tests (including the one from 69f51af, which seems to have landed to v1
only, copying it here) and they all pass.

(As discussed in #311, where it was rediscovered)

### DIFF
--- a/markdown.go
+++ b/markdown.go
@@ -639,9 +639,6 @@ func scanLinkRef(p *parser, data []byte, i int) (linkOffset, linkEnd, titleOffse
 	for i < len(data) && data[i] != ' ' && data[i] != '\t' && data[i] != '\n' && data[i] != '\r' {
 		i++
 	}
-	if i == len(data) {
-		return
-	}
 	linkEnd = i
 	if data[linkOffset] == '<' && data[linkEnd-1] == '>' {
 		linkOffset++
--- a/inline_test.go
+++ b/inline_test.go
@@ -604,6 +604,9 @@ func TestReferenceLink(t *testing.T) {
 
 		"[ref]\n   [ref]: ../url/ \"title\"\n",
 		"<p><a href=\"../url/\" title=\"title\">ref</a></p>\n",
+
+		"[link][ref]\n   [ref]: /url/",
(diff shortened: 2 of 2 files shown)


markdown.go
```diff
@@ -639,9 +639,6 @@ func scanLinkRef(p *parser, data []byte, i int) (linkOffset, linkEnd, titleOffse
 	for i < len(data) && data[i] != ' ' && data[i] != '\t' && data[i] != '\n' && data[i] != '\r' {
 		i++
 	}
-	if i == len(data) {
-		return
-	}
```
shurcooL: Just a thought, but do you know of any preconditions that must be met when calling `scanLinkRef`? If so, it would be tremendously helpful to document them. That way, it may become more viable to verify that the code inside does the right thing on all input.

As it stands now, without documentation, this func can accept arbitrary input, and it's not very clear what the exact responsibility/postconditions this func is expected to produce.

rtfb: Ohh, that's a very good question that I don't have a good answer to. It's only being called from `isReference`, which in turn is only called from the `paragraph` block parser, but that doesn't say much about preconditions. Other than that, it will take some time to study those nearly 200 lines carefully enough to tease out all preconditions. I'll take a note to do that when I have chance.

PR #334, opened 2017-02-22
add an extension to handle Chinese (or CJK) newlines.

choueric:
There is a common discussion, for example [here](http://stackoverflow.com/questions/8550112/prevent-browser-converting-n-between-lines-into-space-for-chinese-characters/8551033#8551033), about the difference between English and Chinese when there are linebreaks in a paragraph. Here is a method by adding a extension to join lines.


rtfb:
Sorry for the delay to get back to you.

First, please rewrite the test to not need the testdata file. I see no reason for that. Also, move the test to `block_test.go`.

Also, how does this work with consecutive newline characters (and how should it)? I see from the test that it doesn't join all the input into a single line, but can't immediately see why.

choueric:
Thanks for your suggestion and I did the modifications.

About the newline, there are actually two paragraphes in the testdata, one is:
```
第一
行文字。
```
and the other is:
```
第
二
行文字。
```
So it should generate two lines:
```
<p>第一行文字。</p>
````
and
```
<p>第二行文字。</p>
```
Consecutive newline characters, for example two newlines, are used to separate two paragraphes and should not generate just one line (i.e. one paragraph).

If I don't get your idea, please forgive me. Or you can give me an example to illustrate.

rtfb:
Thanks, @choueric!

PR #337, opened 2017-03-02
Clean up Options struct and Markdown func

rtfb:
Move renderer from a formal parameter to Markdown to an Options member.

This is an (intentionally conflicting) alternative to #328. Only one of them should land, I'm providing both in order to compare them against each other in terms of actual code.

### DIFF
--- a/markdown.go
+++ b/markdown.go
@@ -59,6 +59,9 @@ const (
 // enabled by default.
 var DefaultOptions = Options{
 	Extensions: CommonExtensions,
+	Renderer: NewHTMLRenderer(HTMLRendererParameters{
+		Flags: CommonHTMLFlags,
+	}),
 }
 
 // ListType contains bitwise or'ed flags for list and list item objects.
@@ -281,18 +284,20 @@ type Options struct {
 	// the override function indicates an override did not occur, the refids at
 	// the bottom will be used to fill in the link details.
 	ReferenceOverride ReferenceOverrideFunc
+
+	// Renderer holds a renderer used to convert the parsed syntax tree to the
+	// desired output form. Blackfriday itself provides one renderer: HTML.
+	Renderer Renderer
 }
 
 // MarkdownBasic is a convenience function for simple rendering.
 // It processes markdown input with no extensions enabled.
 func MarkdownBasic(input []byte) []byte {
-	// set up the HTML renderer
-	renderer := NewHTMLRenderer(HTMLRendererParameters{
-		Flags: UseXHTML,
+	return Markdown(input, Options{
+		Renderer: NewHTMLRenderer(HTMLRendererParameters{
+			Flags: UseXHTML,
+		}),
 	})
-
-	// set up the parser
-	return Markdown(input, renderer, Options{})
 }
 
 // MarkdownCommon is a convenience function for simple rendering. It calls
@@ -314,24 +319,20 @@ func MarkdownBasic(input []byte) []byte {
 //
 // * Custom Heading IDs
 func MarkdownCommon(input []byte) []byte {
-	// set up the HTML renderer
-	renderer := NewHTMLRenderer(HTMLRendererParameters{
-		Flags: CommonHTMLFlags,
-	})
-	return Markdown(input, renderer, DefaultOptions)
+	return Markdown(input, DefaultOptions)
 }
 
 // Markdown is the main rendering function.
 // It parses and renders a block of markdown-encoded text.
-// The supplied Renderer is used to format the output, and extensions dictates
-// which non-standard extensions are enabled.
+// The supplied options contain a Renderer used to format the output, and
+// extensions that dictate which non-standard extensions are enabled.
 //
-// To use the supplied HTML renderer, see NewHTMLRenderer.
-func Markdown(input []byte, renderer Renderer, options Options) []byte {
-	if renderer == nil {
-		return nil
+// If options.Renderer is nil, the supplied HTML renderer is used.
+func Markdown(input []byte, options Options) []byte {
+	if options.Renderer == nil {
+		options.Renderer = DefaultOptions.Renderer
 	}
-	return renderer.Render(Parse(input, options))
+	return options.Renderer.Render(Parse(input, options))
 }
 
 // Parse is an entry point to the parsing part of Blackfriday. It takes an
--- a/markdown.go
+++ b/markdown.go
@@ -56,7 +56,12 @@ const (
 )
 
 // DefaultOptions is a convenience variable with all the options that are
-// enabled by default.
+// enabled by default. Namely, a set of CommonExtensions plus the default
+// HTML renderer configured with a set of CommonHTMLFlags.
+//
+// Do not modify DefaultOptions variable since that might have side effects on
+// later invocations of Markdown* functions. If you need to customize behavior,
+// pass your own copy of Options to the Markdown function.
 var DefaultOptions = Options{
 	Extensions: CommonExtensions,
 	Renderer: NewHTMLRenderer(HTMLRendererParameters{
@@ -301,7 +306,8 @@ func MarkdownBasic(input []byte) []byte {
 }
 
 // MarkdownCommon is a convenience function for simple rendering. It calls
-// Markdown with most useful extensions enabled, including:
+// Markdown with DefaultOptions, which contain the most useful extensions
+// enabled, including:
 //
 // * Smartypants processing with smart fractions and LaTeX dashes
 //
@@ -322,12 +328,12 @@ func MarkdownCommon(input []byte) []byte {
 	return Markdown(input, DefaultOptions)
 }
 
-// Markdown is the main rendering function.
-// It parses and renders a block of markdown-encoded text.
-// The supplied options contain a Renderer used to format the output, and
-// extensions that dictate which non-standard extensions are enabled.
+// Markdown is the main entry point. It parses and renders a block of
+// markdown-formatted text. The supplied options contain a Renderer used to
+// format the output, and extensions that dictate which non-standard extensions
+// are enabled.
 //
-// If options.Renderer is nil, the supplied HTML renderer is used.
+// If options.Renderer is nil, then DefaultOptions.Renderer is used.
 func Markdown(input []byte, options Options) []byte {
 	if options.Renderer == nil {
 		options.Renderer = DefaultOptions.Renderer
--- a/markdown.go
+++ b/markdown.go
@@ -24,45 +24,67 @@ const Version = "2.0"
 
 // Extensions is a bitwise or'ed collection of enabled Blackfriday's
 // extensions.
-type Extensions int
+type Extensions int64
 
 // These are the supported markdown parsing extensions.
 // OR these values together to select multiple extensions.
 const (
-	NoExtensions           Extensions = 0
-	NoIntraEmphasis        Extensions = 1 << iota // Ignore emphasis markers inside words
-	Tables                                        // Render tables
-	FencedCode                                    // Render fenced code blocks
-	Autolink                                      // Detect embedded URLs that are not explicitly marked
-	Strikethrough                                 // Strikethrough text using ~~test~~
-	LaxHTMLBlocks                                 // Loosen up HTML block parsing rules
-	SpaceHeadings                                 // Be strict about prefix heading rules
-	HardLineBreak                                 // Translate newlines into line breaks
-	TabSizeEight                                  // Expand tabs to eight spaces instead of four
-	Footnotes                                     // Pandoc-style footnotes
-	NoEmptyLineBeforeBlock                        // No need to insert an empty line to start a (code, quote, ordered list, unordered list) block
-	HeadingIDs                                    // specify heading IDs  with {#id}
-	Titleblock                                    // Titleblock ala pandoc
-	AutoHeadingIDs                                // Create the heading ID from the text
-	BackslashLineBreak                            // Translate trailing backslashes into line breaks
-	DefinitionLists                               // Render definition lists
-
-	CommonHTMLFlags HTMLFlags = UseXHTML | Smartypants |
-		SmartypantsFractions | SmartypantsDashes | SmartypantsLatexDashes
-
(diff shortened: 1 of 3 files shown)


shurcooL:
> Only one of them should land, I'm providing both in order to compare them against each other in terms of actual code.

That's a great idea and helpful, thanks for doing this.

markdown.go
```diff
@@ -314,24 +319,20 @@ func MarkdownBasic(input []byte) []byte {
...
-// The supplied Renderer is used to format the output, and extensions dictates
-// which non-standard extensions are enabled.
+// The supplied options contain a Renderer used to format the output, and
+// extensions that dictate which non-standard extensions are enabled.
 //
-// To use the supplied HTML renderer, see NewHTMLRenderer.
-func Markdown(input []byte, renderer Renderer, options Options) []byte {
-	if renderer == nil {
-		return nil
+// If options.Renderer is nil, the supplied HTML renderer is used.
```
shurcooL: > If options.Renderer is nil, the supplied HTML renderer is used.

That's not accurate. I think you meant "If options.Renderer is nil, the default HTML renderer is used."

I would even consider writing it more explicitly as:

```
// If options.Renderer is nil, then DefaultOptions.Renderer is used.
```

That's equally readable, more accurate, and takes care of the situation if the user modifies `DefaultOptions`. Should we document that `DefaultOptions` should not be modified, or at least warn that if it is, that has effect on other methods?

markdown.go
```diff
@@ -314,24 +319,20 @@ func MarkdownBasic(input []byte) []byte {
...
-	return Markdown(input, renderer, DefaultOptions)
+	return Markdown(input, DefaultOptions)
 }
 
 // Markdown is the main rendering function.
 // It parses and renders a block of markdown-encoded text.
-// The supplied Renderer is used to format the output, and extensions dictates
-// which non-standard extensions are enabled.
+// The supplied options contain a Renderer used to format the output, and
+// extensions that dictate which non-standard extensions are enabled.
```
shurcooL: This doesn't seem to make for a friendly API, as I understand it, because providing a zero value for blackfriday.Options{} doesn't do what one would expect in this style of API.

Here's the behavior I expect and think we should provide:

```Go
out := blackfriday.Markdown(in, blackfriday.Options{})
```

That should mean "use default options, nothing is overridden to custom values".

Next,

```Go
out := blackfriday.Markdown(in, blackfriday.Options{
	Extensions: extensions,
	Renderer:   renderer,
})
```

That should mean "use the provided extensions and renderer."

shurcooL: Hmm, maybe that's already the case? I guess it depends on what we want "default options" to be, should it be zero extensions, or common extensions?

markdown.go
rtfb: Oh yes, by "supplied" I meant exactly that, the default one. That's an obviously poor choice of words. I like your suggestion to be explicit about what's used instead of nil.

markdown.go
rtfb: Yeah, that's a fundamental problem of trying to provide a set of sane defaults: how do you communicate what the zero does: is it the zero customization or is it the default options?

The current design (and I didn't change it with this PR) intends users to call the convenience functions `MarkdownBasic` and `MarkdownCommon` to distinguish between the zero options and default options. This might, or might not be easy to grasp from the docs; we could explicitly describe  the three invocations in the package-level section.

`DefaultOptions` is another helper to distinguish that, but this one, too, might slip the documentation reader's attention.

But our API surface is very small, maybe we're overthinking this bit :-)

shurcooL:
> Yeah, that's a fundamental problem of trying to provide a set of sane defaults: how do you communicate what the zero does: is it the zero customization or is it the default options?
>
> The current design (and I didn't change it with this PR) intends users to call the convenience functions `MarkdownBasic` and `MarkdownCommon` to distinguish between the zero options and default options. This might, or might not be easy to grasp from the docs; we could explicitly describe  the three invocations in the package-level section.
>
> `DefaultOptions` is another helper to distinguish that, but this one, too, might slip the documentation reader's attention.
>
> But our API surface is very small, maybe we're overthinking this bit :-)

Absolutely, happy to discuss this.

Given this is an experimental PR, we can consider some more drastic changes and see how they look.

### `Options` Struct Variations

As a starting point, I would recommend looking at the approach [`golang.org/x/tools/imports`](https://godoc.org/golang.org/x/tools/imports) package uses for this. It defines an `Options` struct like this (note the comments on struct fields):

```Go
// Options specifies options for processing files.
type Options struct {
	Fragment  bool // Accept fragment of a source file (no package statement)
	AllErrors bool // Report all errors (not just the first 10 on different lines)

	Comments  bool // Print comments (true if nil *Options provided)
	TabIndent bool // Use tabs for indent (true if nil *Options provided)
	TabWidth  int  // Tab width (8 if nil *Options provided)

	FormatOnly bool // Disable the insertion and deletion of imports
}
```

A critical component to that is the definition of [`Process`](), and the fact it takes a pointer to `Options` rather than value:

```
// Process formats and adjusts imports for the provided file. If opt is nil the defaults are used. [...]
func Process(filename string, src []byte, opt *Options) ([]byte, error) {
```

This can be used in two ways.

For a user who doesn't care and just wants to use default options, they can simply do:

```Go
b, err := imports.Process(filename, src, nil) // using defaults
```

If they do care about overriding some options, they can do that, keeping in mind the defaults that are all documented in `Options` struct. For example, if they want to set `Comments` to `false`, they would have to now do:

```Go
b, err := imports.Process(filename, src, &imports.Options{
	Comments:  false, // use explicit desired value instead of default
	TabIndent: true,  // continue to use default value
	TabWidth:  8,     // continue to use default value
})
```

It's a good starting point, but having to keep in mind that some default values are not zero values definitely makes the API slightly harder and more error-prone to use.

One idea to iterate on that is to make the Options struct use optional values, and document that unless provided a different value, a default is used. So let's try changing the above to:

```Go
// Options specifies options for processing files.
type Options struct {
	Fragment  bool // Accept fragment of a source file (no package statement)
	AllErrors bool // Report all errors (not just the first 10 on different lines)

	Comments  *bool // Print comments (true if nil)
	TabIndent *bool // Use tabs for indent (true if nil)
	TabWidth  *int  // Tab width (8 if nil)

	FormatOnly bool // Disable the insertion and deletion of imports
}
```

And make `Process` take `Options` as value. Using default options becomes:

```Go
b, err := imports.Process(filename, src, imports.Options{}) // using defaults
```

And overriding just `Comments` to `false` looks like this:

```Go
comments := false
b, err := imports.Process(filename, src, imports.Options{
	Comments:  &comments, // use explicit desired value instead of default
})
```

Setting fields that are pointers to basic types is definitely awkward, so I think that's the biggest downside of this API. But otherwise this is the best approach I can think of so far.

### Reducing Entry Points in API

One more thing to consider...

As long as we have `Markdown(input []byte, options Options)`, I think we should replace `MarkdownBasic(input []byte)` and `MarkdownCommon(input []byte)` with:

```Go
var BasicOptions = Options{
	// ... actual values for for MarkdownBasic does
}

var CommonOptions = Options{
	// ... actual values for for MarkdownCommn does
}
```

That way there's only one entrypoint, the `Markdown` func that takes options, and the user can do:

```Go
// replaces the current MarkdownBasic
blackfriday.Markdown(input, markdown.BasicOptions)

// replaces the current MarkdownCommon
blackfriday.Markdown(input, markdown.CommonOptions)
```

rtfb:
Ugh. I tried a couple times to give a detailed response here, but I see that all it does is it delays the answer :-). Instead, here's a quick reply in English and I'm off to try and give another reply in Go.

* I like the idea to ditch `MarkdownBasic` and MarkdownCommon`. IMO it improves discoverability.
* I like the idea to make `Options{}` equivalent to `MarkdownCommon`. Having a zero object do the right thing is an idiomatic Go and a good design principle in general, so let's go for it.
* I will define a separate preset for zero/basic options.

rtfb:
Pushed an update. I wonder, can we easily see how the documentation looks in godoc.org w/o merging?

shurcooL:
> Pushed an update.

I'll be able to look at it after Friday.

> I wonder, can we easily see how the documentation looks in godoc.org w/o merging?

That's something I want to add to gotools.org. It already has the ability to access Go packages at arbitrary branches, so I just need to add a tab that displays `godoc`. Until then, using `godoc -http` locally is the only way I know of.

shurcooL:
Sorry for the delay, I had to prepare for a short guest talk on Friday. I've been able to look at this now, and I have further thoughts. Thanks for applying some of the ideas so far, @rtfb.

This is an improvement over before, but I think there's still room for improvement of API.

I just had an idea that I think can enable a lot of that.

For context, look at:

```Go
// Markdown is the main entry point. It parses and renders a block of
// markdown-formatted text. The supplied options contain a Renderer used to
// format the output, and extensions that dictate which non-standard extensions
// are enabled.
//
// If the zero object Options{} is passed for options, CommonExtensions and the
// default HTML renderer are used.
//
// If you want to suppress extensions, pass ZeroOptions.
func Markdown(input []byte, options Options) []byte {
```

Specifically, this part:

```
// If the zero object Options{} is passed for options, CommonExtensions and the
// default HTML renderer are used.
//
// If you want to suppress extensions, pass ZeroOptions.
```

That seems confusing and suboptimal to me.

How about this. I think the ideal API is one where the most commonly done rendering mode looks like this:

```Go
output := blackfriday.Markdown(input, blackfriday.Options{}) // Use defaults, no customization of any kind.
```

However, changing one option should not have a drastic effect on the way that code looks like.

We currently have a lot of extensions and options that are commonly enabled. What if we flip them so that the extension is about "disabling" said option, and zero value means the default behavior of having that extension on.

So, `CommonExtensions` becomes:

```Go
const CommonExtensions Extensions = 0 // default zero value gives common extensions.
```

And `NoExtensions` becomes:

```Go
const NoExtensions = IntraEmphasis | NoTables | NoFencedCode |
    NoAutolink | NoStrikethrough | NoSpaceHeadings | NoHeadingIDs |
    NoBackslashLineBreak | NoDefinitionLists
```

Then there's no need for this hidden "magic":

```Go
// defaultOptions is a single point of truth for default options, to make sure
// the tests and code use the same.
var defaultOptions = Options{
	Extensions: CommonExtensions,
	Renderer: NewHTMLRenderer(HTMLRendererParameters{
		Flags: CommonHTMLFlags,
	}),
}
```

Instead, the zero `Options{}` struct is literally what it takes to use common extension, our default.

Same idea can apply to HTML flags:

```Go
// CommonHTMLFlags defines the common options for the default HTML renderer.
const CommonHTMLFlags HTMLFlags = UseXHTML | Smartypants |
	SmartypantsFractions | SmartypantsDashes | SmartypantsLatexDashes
```

If we change the HTML flags to `UseHTML`, `NoSmartypants`, `NoSmartypantsFractions`, `NoSmartypantsDashes`, `NoSmartypantsLatexDashes`, then:

```
// CommonHTMLFlags defines the common options for the default HTML renderer.
const CommonHTMLFlags HTMLFlags = 0
```

(`CommonHTMLFlags` becomes unnecessary, the above is for demonstration purposes.)

My goal is to make the zero value useful by making it perform default/common rendering behavior, and any extensions/options can be opted-in individually, without a sudden jump in behavior.

Thoughts on this idea of flipping extensions and html flags?

shurcooL:
Now that I've written up that idea, and thought more about it, I'm not completely confident it's a great fit for blackfriday's API. The goal is worthwhile, but it seems a Markdown parser is best thought of as a collection of "features/options" that are included or turned on, rather than a set of features you disable.

But, I really think the following is suboptimal, and I want to find a way to get rid of the need to have this.

```Go
var defaultOptions = Options{
	Extensions: CommonExtensions,
	Renderer: NewHTMLRenderer(HTMLRendererParameters{
		Flags: CommonHTMLFlags,
	}),
}
```

It's hidden logic that is hard for users to understand, and hard to document the API in a readable way. And I don't like the nested nature of the information in `Options`; perhaps if we made the options more flat, they'd be easier to deal with.

rtfb:
Yeah, inverting the flags has crossed my mind too, but I didn't rush with it due to high volume of changes it would requite for a wild experiment. I don't yet have a good idea on how would that feel to the user.

@shurcooL, how would you weigh the functional options now, after the more detailed exploration of Options struct?

Ambrevar:
Aren't we considering the problem upside down?

"Common extensions" are what they are, extensions, or in other words, _non-standard_.

Shouldn't the default leave out every extension? This is what the user would expect from the very meaning of the word "extension."

(This would be much more Unix/KISS/Suckless too!)

Besides, the `Markdown()` function is usually called only once, so in practice this would mean little additional typing:

* `Markdown(input, nil)` for a barren parser.
* `Markdown(input, CommonOptions)` for a "common use" with the common extensions loaded up. It's just a few more characters after all.

I am still slightly in favor of the functional options:

- The following awkwardness would not even occur:
```
if extensions == 0 {
	extensions = defaultOptions.Extensions
}
```
- The  issue with the meaning of the zero-value disappears.

rtfb:
> Shouldn't the default leave out every extension?

Not sure. On the one hand, "common extensions" are what they are -- extensions. But on the other hand, they are what they are -- common :-). I doubt that there are all that much uses of `MarkdownBasic` in the wild. The most easy and obvious way to use Blackfriday should yield you the most common use case.

I'm leaning towards functional options as well.

shurcooL:
> @shurcooL, how would you weigh the functional options now, after the more detailed exploration of Options struct?

I do not have objections to going with functional options. We've explored options struct approach here, and I think it's clear it needs more work, it's not good enough in the current state.

If it were only me working on this, I would still "keep at it" and try to make options struct work better. But I'm not the most active developer here, and I would like to learn more about the functional approach and get a chance to see how it plays out.

So I think you should go with you think is better, @rtfb. I think by now you have a pretty fair comparison of both, so it's not like it'd a be decision made with not enough information. Thanks for considering both approaches this thoroughly!

Ambrevar:
The word _extension_ is both historical and unfortunate: they are now _commonly_ considered as _features_. Not sure we can change that, but maybe that would deserve a comment in the doc.

PR #343, opened 2017-03-18
Add rel=noopener

jhabdas:
For better or worse, browsers have implemented a security and performance feature for links using the `rel=noopener`, preventing the hijacking of the `window.opener.location` object for pages opened with a `_blank` target, which becomes important for sites with UGC.

This commit adds to Blackfriday the ability for users to configure an option to enable the `rel=noopener` attribute.

Refs:

- https://developers.google.com/web/tools/lighthouse/audits/noopener
- https://mathiasbynens.github.io/rel-noopener/
- https://jakearchibald.com/2016/performance-benefits-of-rel-noopener/

### DIFF
--- a/html.go
+++ b/html.go
@@ -32,6 +32,7 @@ const (
 	HTML_SAFELINK                              // only link to trusted protocols
 	HTML_NOFOLLOW_LINKS                        // only link with rel="nofollow"
 	HTML_NOREFERRER_LINKS                      // only link with rel="noreferrer"
+	HTML_NOOPENER_LINKS                        // only link with rel="noopener"
 	HTML_HREF_TARGET_BLANK                     // add a blank target
 	HTML_TOC                                   // generate a table of contents
 	HTML_OMIT_CONTENTS                         // skip the main contents (for a standalone table of contents)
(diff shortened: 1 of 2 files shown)


shurcooL:
I want to point something out for discussion.

Should it be responsibility of blackfriday to do this, given that bluemonday already does this?

In a normal setup where blackfriday is used for rendering user generated content, the output of blackfriday is unsafe HTML which must be sanitized before it's safe to render. A recommended library is bluemonday.

https://github.com/russross/blackfriday#sanitize-untrusted-content

We've had blackfriday _try_ to do HTML sanitization in the past, but decided it was a bad idea. It was done poorly, and it's best to leave that "out of scope" of the renderer.

So, where does rel=noopener fit in here?

See here for information on bluemonday taking care of rel=noopener as part of its sanitization process:

https://github.com/microcosm-cc/bluemonday/issues/35

jhabdas:
@shurcooL good I didn't know that. one thing to keep in mind is that [hugo uses blackfriday wholesale](https://github.com/spf13/hugo/blob/86e8dd62f0b1c8adf14e9369b089d209317aaf2d/helpers/content.go#L30), and there are many people riding on it. would be interested to hear @spf13's thoughts on pulling in the sanitizer as part of Hugo

shurcooL:
There's an open issue https://github.com/spf13/hugo/issues/1457 about that.

jhabdas:
Missed that. Thanks @shurcooL

PR #352, opened 2017-05-02
Document SanitizedAnchorName algorithm, copy implementation.

### LINKED ISSUE #350: copy sanitized_anchor_name to blackfriday
The `blackfriday` package only has one external dependency: `github.com/shurcooL/sanitized_anchor_name`

That repo provides a single ~20-line function. Can we copy it into this repository to remove the external dependency? 

cc @shurcooL

shurcooL:
The goal of this change is to reduce number of non-standard library packages (repositories) that `blackfriday` imports [from 1](https://cloud.githubusercontent.com/assets/1924134/25837387/a192dcb2-345b-11e7-8f1c-1b6a400b8943.png) to 0, and in turn, reduce the cost of importing `blackfriday` into other projects.

Do so by documenting the algorithm of `SanitizedAnchorName`, and include a copy of the small function inside `blackfriday` itself. The same functionality continues to be available in the original location, [`github.com/shurcooL/sanitized_anchor_name.Create`](https://godoc.org/github.com/shurcooL/sanitized_anchor_name#Create). It can be used by existing users and those that look for a small package, and don't need all of `blackfriday` functionality. Existing users of `blackfriday` can use the new `SanitizedAnchorName` function directly and avoid an extra package import.

Resolves #350.


doc.go
```diff
@@ -0,0 +1,32 @@
...
+// as well as invalid character that follow the last valid character
+// are dropped completely. All other sequences of invalid characters
+// between two valid characters are replaced with a single dash character '-'.
+//
+// SanitizedAnchorName exposes this functionality, and can be used to
+// create compatible links to the anchor names generated by blackfriday.
+// This algorithm is also implemented in a small standalone package at
+// github.com/shurcooL/sanitized_anchor_name. It can be useful for clients
+// that want a small package and don't need full functionality of blackfriday.
+package blackfriday
```
shurcooL: Here's what it looks like, for reference.

<details>

image

</details>

russross:
LGTM--Thanks for doing this!

rtfb:
This is great, thanks! The only complaint that I have is the duplicate paragraphs in README and doc.go. Would it perhaps be better to leave the reference text in doc.go and link to it from the README?

shurcooL:
> The only complaint that I have is the duplicate paragraphs in README and doc.go. Would it perhaps be better to leave the reference text in doc.go and link to it from the README?

I did that based on how the [go-github](https://godoc.org/github.com/google/go-github/github) package did it. It has similar documentation in godoc and README:

-	https://godoc.org/github.com/google/go-github/github
-	https://github.com/google/go-github#readme

And we keep them in sync (https://github.com/google/go-github/issues/397).

Personally, I always prefer looking at godoc, but I guess the idea is that beginners to Go may miss that, so putting it in the README helps them catch it.

However, I like your idea of linking to the algorithm specification from the README to the godoc. That way, the specification is in one canonical place and cannot get out of sync.

I'll apply that change.

shurcooL:
@rtfb, how do you feel about adding an exported version of `SanitizedAnchorName` to the `blackfriday` API vs keeping it unexported?

Also, if we do export it, do you have opinions on whether it should be inside `blackfriday` package itself, or if it can be in a small standalone subpackage (inside blackfriday repository)?

rtfb:
> @rtfb, how do you feel about adding an exported version of SanitizedAnchorName to the blackfriday API vs keeping it unexported?

I guess that depends on how likely are people to have projects which call blackfriday in one place and SanitizedAnchorName in another. I can certainly imagine that, but the likelihood seems small. I don't have strong opinion either way, both seem to be adequate.

> Also, if we do export it, do you have opinions on whether it should be inside blackfriday package itself, or if it can be in a small standalone subpackage (inside blackfriday repository)?

I say, inside. Can't foresee that subpackage growing, and one function packages are a bit weird, especially when the function is so specific.

shurcooL:
@adg, how does this solution for #350 look to you?

adg:
LGTM thanks.

PR #356, opened 2017-05-10
Failing test cases for #291

waseem:
Links with parenthesis (like wikipedia links) are not correct when using autolinks

PR #222, opened 2015-11-29
Allow parentheses inside of links

icco:
Fixes #116. Most code originally by @miekg.

Issue #123, opened 2014-10-22
Link parenthesis evaluation seems wrong.


icco:
The following markdown: `[Chef](https://en.wikipedia.org/wiki/Chef_(film))` evaluates to [Chef](https://en.wikipedia.org/wiki/Chef_%28film%29) here on Github, but blackfriday evaluates it to

``` html
<a href="https://en.wikipedia.org/wiki/Chef_(film">Chef</a>)
```

The code I'm using to evaluate this string is:

``` go
func Markdown(args ...interface{}) template.HTML {
  inc := []byte(fmt.Sprintf("%s", args...))
  inc = twitterHandleToMarkdown(inc)
  inc = hashTagsToMarkdown(inc)
  s := blackfriday.MarkdownCommon(inc)
  return template.HTML(s)
}

func twitterHandleToMarkdown(in []byte) []byte {
  return TwitterHandleRegex.ReplaceAll(in, []byte("$1[@$2](http://twitter.com/$2)"))
}

func hashTagsToMarkdown(in []byte) []byte {
  return HashtagRegex.ReplaceAll(in, []byte("$1[#$2](/tags/$2)"))
}
```

Thanks!

cup:
https://github.com/russross/blackfriday/issues/116

https://github.com/russross/blackfriday/issues/123
https://github.com/russross/blackfriday/issues/291
https://github.com/russross/blackfriday/pull/222
https://github.com/russross/blackfriday/pull/356
https://github.com/russross/blackfriday/pull/358

Issue #116, opened 2014-09-16
Link parsing doesn't correctly match URLs when they include brackets


buro9:
For a given piece of Markdown:
`[disambiguation](http://en.wikipedia.org/wiki/Disambiguation_(disambiguation))`

I would expect to see the output:
`<p><a href="http://en.wikipedia.org/wiki/Disambiguation_(disambiguation)">disambiguation</a></p>`

However blackfriday moves the second `)` outside of the link, thus breaking the URL:
`<p><a href="http://en.wikipedia.org/wiki/Disambiguation_(disambiguation">disambiguation</a>)</p>`

You can recreate this with this example code:

```go
package main

import (
	"fmt"

	"github.com/russross/blackfriday"
)

func main() {
	md := []byte(`[disambiguation](http://en.wikipedia.org/wiki/Disambiguation_(disambiguation))`)

	html := blackfriday.MarkdownCommon(md)

	fmt.Println(`Output  : ` + string(html))
	fmt.Println(`Expected: <p><a href="http://en.wikipedia.org/wiki/Disambiguation_(disambiguation)">disambiguation</a></p>`)
}
```

The correct output is generated by most other Markdown parsers, here's Pandoc as the online link makes it the easiest to share:
http://johnmacfarlane.net/pandoc/try/?text=[disambiguation]%28http%3A%2F%2Fen.wikipedia.org%2Fwiki%2FDisambiguation_%28disambiguation%29%29&from=markdown&to=html

Bracketed links are very popular on Wikipedia and a few other top-100 internet sites.

SamuelDebruyn:
@miekg, can you do a PR to blackfriday with your fix?

icco:
Yeah, I run into this almost daily still, right now I just go to http://dev.w3.org/html5/html-author/charref and hand replace parens, but it's not a great fix.

cup:
https://github.com/russross/blackfriday/issues/116
https://github.com/russross/blackfriday/issues/123
https://github.com/russross/blackfriday/issues/291
https://github.com/russross/blackfriday/pull/222
https://github.com/russross/blackfriday/pull/356
https://github.com/russross/blackfriday/pull/358

cup:
https://github.com/mmarkdown/mmark/issues/98

PR #358, opened 2017-05-12
Handle Parenthesis in links properly

### LINKED ISSUE #291: Links with parenthesis (like wikipedia links) are not correct when using autolinks
For example if you have some markdown text with:

```md
Check out this link:
https://en.wikipedia.org/wiki/St._Lucia_(musician)
```

The final markup ends up being

```html
<p>Check out this link:</p>

<p><a href="https://en.wikipedia.org/wiki/St._Lucia_(musician" rel="nofollow">https://en.wikipedia.org/wiki/St._Lucia_(musician</a>)</p>
```

Notice the final closing parenthesis is not matched

waseem:
Links with parenthesis (like wikipedia links) are not correct when using autolinks

Properly handle parenthesis in links.

Fixes #291

### DIFF
--- a/inline.go
+++ b/inline.go
@@ -244,8 +244,6 @@ func link(p *parser, out *bytes.Buffer, data []byte, offset int) int {
 		i++
 	}
 
-	brace := 0
-
 	// look for the matching closing bracket
 	for level := 1; level > 0 && i < len(data); i++ {
 		switch {
@@ -279,6 +277,7 @@ func link(p *parser, out *bytes.Buffer, data []byte, offset int) int {
 		i++
 	}
 
+	parenthesis := 0
 	switch {
 	// inline style link
 	case i < len(data) && data[i] == '(':
@@ -291,7 +290,7 @@ func link(p *parser, out *bytes.Buffer, data []byte, offset int) int {
 
 		linkB := i
 
-		// look for link end: ' " ), check for new opening braces and take this
+		// look for link end: ' " ), check for new opening parenthesis and take this
 		// into account, this may lead for overshooting and probably will require
 		// some fine-tuning.
 	findlinkend:
@@ -301,14 +300,14 @@ func link(p *parser, out *bytes.Buffer, data []byte, offset int) int {
 				i += 2
 
 			case data[i] == '(':
-				brace++
+				parenthesis++
 				i++
 
 			case data[i] == ')':
-				if brace <= 0 {
+				if parenthesis <= 0 {
 					break findlinkend
 				}
-				brace--
+				parenthesis--
 				i++
 
 			case data[i] == '\'' || data[i] == '"':
@@ -793,10 +792,6 @@ func autoLink(p *parser, out *bytes.Buffer, data []byte, offset int) int {
 			}
 
 			bufEnd--
-		}
-
-		if openDelim == 0 {
-			linkEnd--
 		}
 	}
 
--- a/inline_test.go
+++ b/inline_test.go
@@ -570,13 +570,16 @@ func TestInlineLink(t *testing.T) {
 
 		// no closing ), autolinking detects the url next
 		"[disambiguation](http://en.wikipedia.org/wiki/Disambiguation_(disambiguation) is the",
-		"<p>[disambiguation](<a href=\"http://en.wikipedia.org/wiki/Disambiguation_(disambiguation\">http://en.wikipedia.org/wiki/Disambiguation_(disambiguation</a>) is the</p>\n",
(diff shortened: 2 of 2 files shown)


inline_test.go
```diff
@@ -570,13 +570,16 @@ func TestInlineLink(t *testing.T) {
 
 		// no closing ), autolinking detects the url next
 		"[disambiguation](http://en.wikipedia.org/wiki/Disambiguation_(disambiguation) is the",
-		"<p>[disambiguation](<a href=\"http://en.wikipedia.org/wiki/Disambiguation_(disambiguation\">http://en.wikipedia.org/wiki/Disambiguation_(disambiguation</a>) is the</p>\n",
+		"<p>[disambiguation](<a href=\"http://en.wikipedia.org/wiki/Disambiguation_(disambiguation)\">http://en.wikipedia.org/wiki/Disambiguation_(disambiguation)</a> is the</p>\n",
```
waseem: It changes the way unbalanced parenthesis is handled in links. This results in forming a correct link by giving priority to the parenthesis in the link. This however, does not conform(like earlier) to [common mark processing](http://spec.commonmark.org/dingus/?text=foo%20%5Bdisambiguation%5D(http%3A%2F%2Fen.wikipedia.org%2Fwiki%2FDisambiguation_(disambiguation)%20is%20the).

I was trying to make it follow the common mark spec but I could get my way around the code. 🤔

rtfb: Which section of Common Mark spec you had in mind here? It would be great to have this fix conform to Common Mark right away, let's try to see what it takes.

waseem: See: http://spec.commonmark.org/0.27/#link-destination

> A link destination consists of either

>  * a sequence of zero or more characters between an opening < and a closing > that contains no spaces, line breaks, or unescaped < or > characters, or
> * a nonempty sequence of characters that does not include ASCII space or control characters, and includes parentheses only if (a) they are backslash-escaped or (b) they are part of a balanced pair of unescaped parentheses that is not itself inside a balanced pair of unescaped parentheses.

The spec does not talk about unescaped and unbalenced parenthesis in link destination. So I was just trying to make it behave the way http://spec.commonmark.org/dingus/ behaves for such links. It ignores the entire link processing in that case and renders it literally after processing. So:

`foo [disambiguation](http://en.wikipedia.org/wiki/Disambiguation_(disambiguation) is the`

becomes:

`<p>foo [disambiguation](http://en.wikipedia.org/wiki/Disambiguation_(disambiguation) is the</p>`.

waseem: By not talking about unescaped and unbalanced parenthesis, I suppose, the commonmark spec regards those as invalid links and hence does not renders them as HTML links.

inline.go
```diff
@@ -794,10 +793,6 @@ func autoLink(p *parser, out *bytes.Buffer, data []byte, offset int) int {
 
 			bufEnd--
 		}
-
-		if openDelim == 0 {
-			linkEnd--
-		}
```
rtfb: Won't this break this case? `foo (http://en.wikipedia.org/wiki/Disambiguation_(disambiguation)) bar`

Autolink detector should recognize that the outer parentheses do not belong to the link. Can you add this case to the tests?

inline_test.go
rtfb: Looks like the section you're referring to talks about the "fully qualified" links, the ones with full link syntax.

And it [seems](http://spec.commonmark.org/0.27/#autolinks) that CommonMark only has autolinks with a hint: they require angle brackets around the link, which avoids the balancing problem altogether.

cup:
https://github.com/russross/blackfriday/issues/116
https://github.com/russross/blackfriday/issues/123
https://github.com/russross/blackfriday/issues/291
https://github.com/russross/blackfriday/pull/222
https://github.com/russross/blackfriday/pull/356

https://github.com/russross/blackfriday/pull/358

cup:
https://github.com/mmarkdown/mmark/issues/98

cup:
https://github.com/gomarkdown/markdown/issues/134

cup:
Since this pull is being ignored, here is a workaround:

~~~go
package main
import (
   "bytes"
   "github.com/yuin/goldmark"
   "github.com/yuin/goldmark/extension"
)
func main() {
   s1 := "https://en.wikipedia.org/wiki/The_Master_(2012_film)"
   o1 := goldmark.New(goldmark.WithExtensions(extension.Linkify))
   var b2 bytes.Buffer
   o1.Convert([]byte(s1), &b2)
   s2 := b2.String()
   print(s2)
}
~~~

https://github.com/yuin/goldmark

cup:
https://github.com/gohugoio/hugo/issues/5963
https://github.com/mmarkdown/mmark/issues/98

https://github.com/russross/blackfriday/pull/358

PR #359, opened 2017-05-17
Fix role attribute

miekg:


miekg:
Sorry, wrong *repo*. My mmark fork still default to this repo.

PR #365, opened 2017-06-01
Merge Parser into Processor

Ambrevar:


markdown.go
```diff
@@ -343,45 +343,45 @@ func WithNoExtensions() Option {
...
 //         WithRenderer(yourRenderer))
-func Markdown(input []byte, opts ...Option) []byte {
+func Run(input []byte, opts ...Option) []byte {
 	r := NewHTMLRenderer(HTMLRendererParameters{
 		Flags: CommonHTMLFlags,
 	})
 	optList := []Option{WithRenderer(r), WithExtensions(CommonExtensions)}
 	optList = append(optList, opts...)
-	parser := NewProcessor(optList...)
+	parser := New(optList...)
```
rtfb: `New(WithRenderer(r), WithExtensions(CommonExtensions), opts...)` seems more readable, plus avoids a temporary object.

Ambrevar: This is sadly not valid Go: mixing slices and scalars is not allowed for variadic parameters.
https://golang.org/ref/spec#Passing_arguments_to_..._parameters

rtfb: Oh. That's unfortunate... Well, looks good to me, then.

Ambrevar: I believe the rationale behind this design choice was "Minimize implicit allocations."  If your suggested syntax had been allowed Go would have had to allocate a slice behind the scenes.

PR #366, opened 2017-06-08
fix duplicate and recursive footnotes. (#241)

choueric:
Fix the infinite loop when there is a self-refer footnote by checking if
the reference is already added into parser.notes.

It also fixes of creating duplicate footnotes through this modification.

Add coresponding testcase.

### DIFF
--- a/markdown.go
+++ b/markdown.go
@@ -218,7 +218,8 @@ type parser struct {
 	// Footnotes need to be ordered as well as available to quickly check for
 	// presence. If a ref is also a footnote, it's stored both in refs and here
 	// in notes. Slice is nil if footnotes not enabled.
-	notes []*reference
+	notes       []*reference
+	notesRecord map[string]bool
 }
 
 func (p *parser) getRef(refid string) (ref *reference, found bool) {
@@ -242,13 +243,8 @@ func (p *parser) getRef(refid string) (ref *reference, found bool) {
 }
 
 func (p *parser) isFootnote(ref *reference) bool {
-	for _, v := range p.notes {
-		if string(ref.link) == string(v.link) {
-			return true
-		}
-	}
-
-	return false
+	_, ok := p.notesRecord[string(ref.link)]
+	return ok
 }
 
 //
@@ -386,6 +382,7 @@ func MarkdownOptions(input []byte, renderer Renderer, opts Options) []byte {
 
 	if extensions&EXTENSION_FOOTNOTES != 0 {
 		p.notes = make([]*reference, 0)
+		p.notesRecord = make(map[string]bool)
 	}
 
 	first := firstPass(p, input)
--- a/markdown.go
+++ b/markdown.go
@@ -241,6 +241,16 @@ func (p *parser) getRef(refid string) (ref *reference, found bool) {
 	return ref, found
 }
 
+func (p *parser) isFootnote(ref *reference) bool {
+	for _, v := range p.notes {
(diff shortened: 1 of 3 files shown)


markdown.go
```diff
@@ -241,6 +241,16 @@ func (p *parser) getRef(refid string) (ref *reference, found bool) {
 	return ref, found
 }
 
+func (p *parser) isFootnote(ref *reference) bool {
+	for _, v := range p.notes {
```
rtfb: I guess in order to do that we need to change `p.notes` from a slice to a map in order to avoid `n^2` complexity.

choueric: You are right. I thought the two string cnversions may cost too much. A map is a good way to deal this problem. I will try to do this.

arikroc:
@choueric: "I checked the CI log and it is a failure of runing too long (10mins) for golang v1.2.2. The times of other versions are less than 10 minutes."

@rtfb Do you agree?

rtfb:
Restarted, let's give it another spin.

markdown.go
```diff
@@ -218,7 +218,8 @@ type parser struct {
 	// Footnotes need to be ordered as well as available to quickly check for
 	// presence. If a ref is also a footnote, it's stored both in refs and here
 	// in notes. Slice is nil if footnotes not enabled.
-	notes []*reference
+	notes       []*reference
+	notesRecord map[string]bool
```
rtfb: `map[string]struct{}` is a better way to represent a set. It avoids allocating memory for the value.

But it also looks like we already have a map that could be used for this: `p.refs`. Can you check if it could do the job?

choueric: 1. About the `map[string]struct{}`, my understanding is that we can not use a map to represent footnotes because they have to be ordered, which means a slice is the best way like it is now `p.notes`. Another map variable `notesRecord` here is just to make the check of duplicate faster.

2. About `p.refs`, do you mean to use `p.refs` to check the duplicate? I do not think it works. As the comment above the definition of `notes []*refrence` says, `p.refs` represents all the references including image, footnote, url etc., while `p.notes` only contains _valid_ and sorted footnotes. It only makes sense to check `p.notes` because it is used to generate footer.

For example, when the input is
```
This uses footnote A.[^A]
                                                                                                                           
This uses footnote C.[^C]                                                                                                                                   
[^A]:  A note. use itself.[^A]                                                       
[^B]:  B note, uses A to test duplicate.[^A]                                         
[^C]:  C note, uses B.[^B]
[^D]: D note.
```
`p.refs` contains all four footnotes. But `p.notes` only contains A and C before parsing footers. And then when parse footnote C, the ref B is already in `p.refs` but not in `p.notes`. B should be added into `p.notes`. But if according to `p.refs`, it is the duplicate ref and not be added into `p.notes`, which is a wrong behavior.

rtfb: I see. If we can not reuse `p.refs`, I'm fine with another map.

Regarding bullet 1, I meant to use `map[string]struct{}` instead of `map[string]bool`. Notice how in `isFootnote` you discard the value in the map: `_, ok := p.notesRecord[string(ref.link)]`. Instead of holding the bool that is discarded, you can hold a zero-size empty struct and thus use the map as a set:

declaration: `notesSet map[string]struct{}`
assignment: `p.notesSet[string(lr.link)] = struct{}{}`
and the check remains unchanged: `_, ok := p.notesRecord[string(ref.link)]`

rtfb:
Thank you, @choueric!

Issue #239, opened 2016-02-12
Fenced code block with list fails


moorereason:
Given the following fenced code block,

`````` md
1. Foo

    ```
    + bar
    ```
``````

blackfriday renders:

`````` text
<ol>\n<li><p>Foo</p>\n\n<p>```</p>\n\n<ul>\n<li>bar\n```</li>\n</ul></li>\n</ol>\n
``````

I expect it to render:

``` text
<ol>\n<li><p>Foo</p>\n\n<pre><code>+ bar\n</code></pre></li>\n</ol>\n
```

It appears to occur when the first line of the code block matches a list item pattern (ie. `-`, `+`, `*`, or `1.`).  If there's a blank line or any non-list item patterned content on the first line, the code fencing works properly.

Here's a test case for `TestFenceCodeBlockWithinList`:

`````` text
        "1. Foo\n\n    ```\n    + bar\n    ```\n",
        `<ol>
<li><p>Foo</p>

<pre><code>+ bar
</code></pre></li>
</ol>
`,
``````

I haven't been able to figure out where it's breaking down.

**Comment by dynamind** [comment]
I encountered this issue with Jenkins console output, which contains xtrace lines (i.e. shell commands are logged to stdout prefixed by `+` followed by a space). 

Notably, the `+` character was not on the first line. I've taken the liberty to include the whole thing below.

I suspect the code scanning for the `+ ` pattern inside lists simply runs before the code that detects fenced code blocks, thereby causing the rendering issue.

```
Started by user anonymous
[Pipeline] Allocate node : Start
Running on master in /var/jenkins_home/jobs/simple-pipeline/workspace
[Pipeline] node {
[Pipeline] sh
[workspace] Running shell script
+ echo 
            echo "Hello from script"
[Pipeline] load: Loaded script: script.groovy
[Pipeline] load {
[Pipeline] echo
Hello from script
[Pipeline] } //load
[Pipeline] } //node
[Pipeline] Allocate node : End
[Pipeline] End of Pipeline
Finished: SUCCESS
```

**Comment by shurcooL** [comment]
Confirmed, this is a bug.

**Comment by slohse** [comment]
I encountered this problem today due to a line in a code section that starts with ": ".

Disclaimer: this morning I knew nothing about the go programming language.
Nonetheless I looked into it a little. I built a test case in `TestFenceCodeBlockWithinList` very similar to the one @moorereason posted.
```
--- FAIL: TestFencedCodeBlockWithinList (0.00s)
[...]

PR #372, opened 2017-06-11
Fix for issue russross/blackfriday#239

client9:
Fix for issue russross/blackfriday#239

This appears to also work in V1, although this PR is for V2

The following would confuse the `- codeline` as if it were a start of a nested list.

```
- line

    ```
    - codeline
    ```
```

The PR checks if a code fence is found, and don't use code to figure out if a list item has ended.

a code review is most welcome as I'm sure I'm missing something.

thanks!

n

### DIFF
--- a/block.go
+++ b/block.go
@@ -1246,6 +1246,7 @@ func (p *Markdown) listItem(data []byte, flags *ListType) int {
 	// process the following lines
 	containsBlankLine := false
 	sublist := 0
+	codeBlockMarker := ""
 
 gatherlines:
 	for line < len(data) {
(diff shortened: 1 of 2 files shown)


rtfb:
Looks good. How does it work with an unclosed code block? Like this:

```
- line

    ```
    - codeline
```

That's a malformed markdown, of course, and any behavior is almost as good as any other, but I want to make sure no source text gets omitted.

client9:
I'll check... also, are there any issues with markdown parsing options, so that this needs to be turned on or off?

rtfb:
> are there any issues with markdown parsing options, so that this needs to be turned on or off?

Oh I'm glad you asked, because it slipped my attention. Looks like your fix should be conditional and do nothing with `FencedCode` turned off. With no fenced code, that seems to be a valid sublist.

And while you're at it, can you please add a test case for a non-fenced code block with sublist-like content?

kaushalmodi:
Looking forward to this getting merged. Thanks for working on the fix!

tfogo:
Hey @client9 are you planning on continuing work on this fix? It would be fantastic to have #239 fixed.

client9:
@rtfb please re-view.  Thanks to excellent work by @tfogo , we have the tests and conditional that you requested.

rtfb:
LGTM, thanks a bunch, @client9 and @tfogo!!

tfogo:
Awesome thanks @rtfb! Thanks @client9 for the heavy lifting!

PR #376, opened 2017-07-09
v2: improve Renderer and fix #368

### LINKED ISSUE #368: v2: API Enhancement: Why is RenderNode part of the Renderer interface?
Hello v2 question.

This is the interface for a `Renderer` is:

```
type Renderer interface {
	Render(ast *Node) []byte
	RenderNode(w io.Writer, node *Node, entering bool) WalkStatus
}
```

Why is `RenderNode` part of the interface?  I don't see any direct usage of it.  I do see `Render` called.  It's not clear to me why implementers of  a new renderer need to use this method name or signature (and make it public).   But perhaps I missed something, as `node.Walk` and `RenderNode` are similar.

thanks!

n

rtfb:
Improve Renderer to be less confusing. Fix documentation for it.

OmitContents flag got dropped along the way. First, it would fit poorly
into the new design and second, it's unclear how widely this feature is
used. But most importantly, it's trivial to roll your own with the v2
API: https://gist.github.com/rtfb/2693f6bfcc1760661e8d2fb832763a15

Fixes #368.

### DIFF
--- a/html.go
+++ b/html.go
@@ -45,7 +45,6 @@ const (
 	SmartypantsLatexDashes                        // Enable LaTeX-style dashes (with Smartypants)
 	SmartypantsAngledQuotes                       // Enable angled double quotes (with Smartypants) for double quotes rendering
 	TOC                                           // Generate a table of contents
-	OmitContents                                  // Skip the main contents (for a standalone table of contents)
 
 	TagName               = "[A-Za-z][A-Za-z0-9-]*"
 	AttributeName         = "[a-zA-Z_:][a-zA-Z0-9:._-]*"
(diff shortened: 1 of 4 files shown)


html.go
```diff
@@ -819,55 +818,71 @@ func (r *HTMLRenderer) RenderNode(w io.Writer, node *Node, entering bool) WalkSt
...
-		w.WriteString(ending)
-		w.WriteString(">\n")
+		w.Write([]byte("\""))
+		w.Write([]byte(ending))
+		w.Write([]byte(">\n"))
 	}
-	w.WriteString("</head>\n")
-	w.WriteString("<body>\n\n")
+	w.Write([]byte("</head>\n"))
+	w.Write([]byte("<body>\n\n"))
```
shurcooL: I see you had to change these calls from `WriteString` to `Write` because the parameter is now `io.Writer` instead of the previous `*bytes.Buffer`.

Question, have you considered using [`io.WriteString`](https://godoc.org/io#WriteString) instead? E.g.:

```diff
-w.WriteString("<body>\n\n")
+io.WriteString(w, "<body>\n\n")
```

How do you think it compares?

rtfb: Ah, I keep forgetting this thing exists... I find it more readable and changed to it. Thanks!

PR #379, opened 2017-07-22
Add Smartypants support for French Guillemets

### LINKED ISSUE #378: Add spaces to angle quotes for French guillemets
In 2014 support was added for angle quotes in BlackFriday thanks to @bep (#128). However, the French [guillemets](https://en.wikipedia.org/wiki/Guillemet) would also require non-breaking spaces: "this" should not become «that», but « that ».

Such further enhancement would make a great difference for French users, as mentioned in [Hugo issue #3725](https://github.com/gohugoio/hugo/issues/3725).

bep:
This commits adds flag `HTML_SMARTYPANTS_QUOTES_NBSP` which, when combined with `HTML_USE_SMARTYPANTS` will insert non-breaking spaces between the double quotes and the contained text.

This is mostly relevant for use in French  with `HTML_SMARTYPANTS_ANGLED_QUOTES`.

It should not hurt existing code path in the performance department:

```
name                     old time/op    new time/op    delta
SmartDoubleQuotes-4    2.58µs ± 1%    2.58µs ± 1%   ~             (p=1.000 n=5+5)

name                     old alloc/op   new alloc/op   delta
SmartDoubleQuotes-4    5.27kB ± 0%    5.27kB ± 0%   ~     (all samples are equal)

name                     old allocs/op  new allocs/op  delta
SmartDoubleQuotes-4      13.0 ± 0%      13.0 ± 0%   ~     (all samples are equal)
```

Fixes #378

### DIFF
--- a/smartypants.go
+++ b/smartypants.go
@@ -39,7 +39,7 @@ func isdigit(c byte) bool {
 	return c >= '0' && c <= '9'
 }
 
-func smartQuoteHelper(out *bytes.Buffer, previousChar byte, nextChar byte, quote byte, isOpen *bool) bool {
+func smartQuoteHelper(out *bytes.Buffer, previousChar byte, nextChar byte, quote byte, isOpen *bool, addNBSP bool) bool {
 	// edge of the buffer is likely to be a tag that we don't get to see,
 	// so we treat it like text sometimes
 
@@ -96,6 +96,12 @@ func smartQuoteHelper(out *bytes.Buffer, previousChar byte, nextChar byte, quote
 		*isOpen = false
 	}
 
+	// Note that with the limited lookahead, this non-breaking
+	// space will also be appended to single double quotes.
+	if addNBSP && !*isOpen {
+		out.WriteString("&nbsp;")
+	}
+
 	out.WriteByte('&')
 	if *isOpen {
 		out.WriteByte('l')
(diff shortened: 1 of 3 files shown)


rtfb:
The fix looks good. Would you also be willing to forward port it to v2? The internals of Smartypants haven't changed much there, IIRC.

bep:
>The fix looks good. Would you also be willing to forward port it to v2?

No problem, I will do that in a separate PR. I created #380 to track/remind me of it.

smartypants.go
```diff
@@ -367,14 +381,30 @@ type smartCallback func(out *bytes.Buffer, smrt *smartypantsData, previousChar b
...
+var (
+	smartAmpAngled      = smartAmp(true, false)
+	smartAmpAngledNbsp  = smartAmp(true, true)
+	smartAmpRegular     = smartAmp(false, false)
+	smartAmpRegularNbsp = smartAmp(false, true)
+)
+
 func smartypants(flags int) *smartypantsRenderer {
 	r := new(smartypantsRenderer)
+	addNbsp := flags&HTML_SMARTYPANTS_QUOTES_NBSP != 0
```
shurcooL: Style nit, `addNBSP`, see https://golang.org/s/style#initialisms.

bep: Thanks, fixed.

PR #381, opened 2017-07-29
Unexport a bunch of constants private to HTML renderer

rtfb:
They were cluttering the documentation and are not useful otherwise.

### DIFF
--- a/html.go
+++ b/html.go
@@ -47,27 +47,27 @@ const (
 	TOC                                           // Generate a table of contents
 )
 
-const (
-	tagName               = "[A-Za-z][A-Za-z0-9-]*"
-	attributeName         = "[a-zA-Z_:][a-zA-Z0-9:._-]*"
-	unquotedValue         = "[^\"'=<>`\\x00-\\x20]+"
-	singleQuotedValue     = "'[^']*'"
-	doubleQuotedValue     = "\"[^\"]*\""
-	attributeValue        = "(?:" + unquotedValue + "|" + singleQuotedValue + "|" + doubleQuotedValue + ")"
-	attributeValueSpec    = "(?:" + "\\s*=" + "\\s*" + attributeValue + ")"
-	attribute             = "(?:" + "\\s+" + attributeName + attributeValueSpec + "?)"
-	openTag               = "<" + tagName + attribute + "*" + "\\s*/?>"
-	closeTag              = "</" + tagName + "\\s*[>]"
-	htmlComment           = "<!---->|<!--(?:-?[^>-])(?:-?[^-])*-->"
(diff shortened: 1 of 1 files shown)


rtfb:
They're only used to initialize the `htmlTagRe` variable, last line of the diff. That's all they are, a big regexp broken into smaller pieces. The regexp is then used to wipe all html inside ``.

shurcooL:
> They're only used to initialize the `htmlTagRe` variable, last line of the diff.

Ah, I see that now, thanks. 👍

FWIW, I think it's more readable in general to define higher level things on top (which use lower level things). Then define the lower level things below them. That way, the reader sees uses of things that aren't defined yet, but they acquire context (e.g., "ok, so htmlTag is something made up from openTag, closeTag, etc.; now let me see what those things are"). Then they can look up the implementation details of the lower level helpers (if they want to learn more details).

This is possible in Go for package scope identifiers, because they can be arranged in any order and the compiler figures out the correct order to initialize them (based on use). It's harder in languages like C++ where things are defined in order they show up in source code, so people more often define lower level things first, higher level ones later (although headers let you get around that limitation).

So, I think the following arrangement would be more readable:

```Go
const (
	htmlTag               = "(?:" + openTag + "|" + closeTag + "|" + htmlComment + "|" +
		processingInstruction + "|" + declaration + "|" + cdata + ")"
	openTag               = "<" + tagName + attribute + "*" + "\\s*/?>"
	closeTag              = "</" + tagName + "\\s*[>]"
	htmlComment           = "<!---->|<!--(?:-?[^>-])(?:-?[^-])*-->"
	processingInstruction = "[<][?].*?[?][>]"
	declaration           = "<![A-Z]+" + "\\s+[^>]*>"
	cdata                 = "<!\\[CDATA\\[[\\s\\S]*?\\]\\]>"
	tagName               = "[A-Za-z][A-Za-z0-9-]*"
	attribute             = "(?:" + "\\s+" + attributeName + attributeValueSpec + "?)"
	attributeName         = "[a-zA-Z_:][a-zA-Z0-9:._-]*"
	attributeValueSpec    = "(?:" + "\\s*=" + "\\s*" + attributeValue + ")"
	attributeValue        = "(?:" + unquotedValue + "|" + singleQuotedValue + "|" + doubleQuotedValue + ")"
	unquotedValue         = "[^\"'=<>`\\x00-\\x20]+"
	singleQuotedValue     = "'[^']*'"
	doubleQuotedValue     = "\"[^\"]*\""
)
```

This is just an optional style suggestion.

rtfb:
Yeah, this reasoning makes perfect sense. I've rearranged the constants in roughly opposite order.

PR #382, opened 2017-07-29
Several small documentation fixes

rtfb:
I went through README and the docs and picked out some remaining inaccuracies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,16 +62,16 @@ Usage
 For the most sensible markdown processing, it is as simple as getting your input
 into a byte slice and calling:
 
-``` go
-    output := blackfriday.Run(input)
+```go
+output := blackfriday.Run(input)
 ```
 
 Your input will be parsed and the output rendered with a set of most popular
 extensions enabled. If you want the most basic feature set, corresponding with
 the bare Markdown specification, use:
 
-``` go
-    output := blackfriday.Run(input, blackfriday.WithNoExtensions())
+```go
+output := blackfriday.Run(input, blackfriday.WithNoExtensions())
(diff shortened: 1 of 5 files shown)


html.go
```diff
@@ -482,8 +482,8 @@ func (r *HTMLRenderer) outHRTag(w io.Writer) {
 }
 
 // RenderNode is a default renderer of a single node of a syntax tree. For
-// block nodes it will be called twice: first time with entering=true, second
-// time with entering=false, so that it could know when it's working on an open
+// block nodes it will be called twice: first time with entering=True, second
+// time with entering=False, so that it could know when it's working on an open
```
shurcooL: Why this change? Isn't lower case correct? https://godoc.org/builtin#true

README.md
```diff
@@ -52,22 +52,27 @@ Potential drawbacks:
...
 
 Usage
 -----
 
 For the most sensible markdown processing, it is as simple as getting your input
 into a byte slice and calling:
 
+``` go
     output := blackfriday.Run(input)
+```
```
shurcooL: When you go from indented code block to fenced code block, you should get rid of the indentation, since it'll now be a part of the code block. This should become:

    ```go
    output := blackfriday.Run(input)
    ```

Also, it's more common not to include a space after the three backticks.

html.go
rtfb: Wow, that's a brain fart. Obviously too much Python lately :-). I'll fix it, thanks!

README.md
rtfb: Will fix, thanks.

PR #384, opened 2017-07-31
Add Smartypants support for French Guillemets

### LINKED ISSUE #380: Port French Guillemets fix to v2
See #379

bep:
This is a port of the fix for #378 in v1.

There is a slight alloc hit in the `SPRenderer` create, and I'm currently not seeing a smart way to avoid that.

```
name                 old time/op    new time/op    delta
EscapeHTML-4           1.22µs ± 2%    1.21µs ± 1%     ~     (p=0.222 n=5+5)
SmartDoubleQuotes-4    5.42µs ± 1%    5.56µs ± 1%   +2.76%  (p=0.008 n=5+5)
Reference-4            1.22ms ± 1%    1.21ms ± 2%     ~     (p=0.056 n=5+5)

name                 old alloc/op   new alloc/op   delta
EscapeHTML-4            0.00B          0.00B          ~     (all equal)
SmartDoubleQuotes-4    9.65kB ± 0%    9.87kB ± 0%   +2.32%  (p=0.008 n=5+5)
Reference-4            1.15MB ± 0%    1.15MB ± 0%   +0.43%  (p=0.008 n=5+5)

name                 old allocs/op  new allocs/op  delta
EscapeHTML-4             0.00           0.00          ~     (all equal)
SmartDoubleQuotes-4      50.0 ± 0%      56.0 ± 0%  +12.00%  (p=0.008 n=5+5)
Reference-4             4.27k ± 0%     4.40k ± 0%   +3.09%  (p=0.008 n=5+5)
```

Fixes #380

### DIFF
--- a/html.go
+++ b/html.go
@@ -44,6 +44,7 @@ const (
 	SmartypantsDashes                             // Enable smart dashes (with Smartypants)
 	SmartypantsLatexDashes                        // Enable LaTeX-style dashes (with Smartypants)
 	SmartypantsAngledQuotes                       // Enable angled double quotes (with Smartypants) for double quotes rendering
+	SmartypantsQuotesNBSP                         // Enable « French guillemets » (with Smartypants)
 	TOC                                           // Generate a table of contents
 
 	TagName               = "[A-Za-z][A-Za-z0-9-]*"
--- a/inline_test.go
+++ b/inline_test.go
@@ -1034,6 +1034,18 @@ func TestSmartDoubleQuotes(t *testing.T) {
 	doTestsInlineParam(t, tests, TestParams{HTMLFlags: Smartypants})
 }
 
+func TestSmartDoubleQuotesNBSP(t *testing.T) {
+	var tests = []string{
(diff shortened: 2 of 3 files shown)


html.go
```diff
@@ -44,6 +44,7 @@ const (
 	SmartypantsDashes                             // Enable smart dashes (with Smartypants)
 	SmartypantsLatexDashes                        // Enable LaTeX-style dashes (with Smartypants)
 	SmartypantsAngledQuotes                       // Enable angled double quotes (with Smartypants) for double quotes rendering
+	SmartypantsQuotesNBSP                         // Enable "French guillemets" (with Smartypants)
```
shurcooL: Would it be a good idea to use French guillemets here, instead of double quotes? Go source code (and docs) are okay with unicode.

bep: Are you referencing the comment? Yes, that would be a good idea.

shurcooL: Yes, I am. Great!

rtfb:
Thanks a lot, @bep!

PR #398, opened 2017-10-06
Allow EXTENSION_HEADER_IDS {#id} also in alternative headers syntax.

bibainet:
```
Header text   {#id}
-------------------
```

### DIFF
--- a/block.go
+++ b/block.go
@@ -1352,6 +1352,20 @@ func (p *parser) paragraph(out *bytes.Buffer, data []byte) int {
 					eol--
 				}
 
+				id := ""
+				if p.flags&EXTENSION_HEADER_IDS != 0 && eol > prev + 2 && data[eol-1] == '}' {
+					for j := eol - 2; j > prev; j-- {
+						if data[j] == '#' && data[j-1] == '{' {
+							id = string(data[j+1:eol-1])
+							eol = j - 1
+							for eol > prev && (data[eol-1] == ' ' || data[eol-1] == '\t') {
+								eol--
+							}
+							break
+						}
+					}
+				}
+
 				// render the header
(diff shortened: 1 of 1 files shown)


rtfb:
Hi @bibainet,

Sorry for the delay to take a look... There are some problems with this PR, some easy to address, some tougher.

Two easy ones (they're somewhat related): there are no tests, and that makes it difficult to understand what some of the commits are meant to accomplish. For example the first commit, `Allow EXTENSION_HEADER_IDS {#id} also in alternative headers syntax.` -- I can't decipher what it does. The tests would at least make it obvious what's the use case, and I could work from there.

A somewhat tougher problem is the fact that this PR is against master, which is v1 of Blackfriday, which is essentially in maintenance mode and should not get new features. Please take a look at the [`v2`](https://github.com/russross/blackfriday/tree/v2) branch, this is where active development is meant to happen.

And finally, the toughest problem is that some of the commits introduce changes to the way the source document text is interpreted. This proliferates extension zoo, and we run a serious risk of those extensions acting against one another and (even worse) base functionality. I'm trying to keep a complete halt on such extensions until we adopt CommonMark, which at least has a very extensive set of tests with good coverage of corner cases.

Having said that, I will close this PR, it shouldn't land. But I will be happy to help you move some of these commits to v2, and for those that wouldn't go to v2, let's try to find a solution that could live as an external extension to Blackfriday, in your own repo. Does that work for you?

If you'd wish to continue with the above, please file a separate issue for each extension that you propose, complete with sample markdown inputs and HTML outputs, and let's discuss each separately.

bibainet:
Hi @rtfb ,

I will return to this later. Now I do not have time to do it.

rtfb:
Yup, no problem! :-)

PR #400, opened 2017-10-11
Add a link to bfchroma, glue library with Chroma

### LINKED ISSUE #10: Code highlighting
It would be nice to be able to highlight code on server-side instead of doing that on client-side (right now I'm thinking about HTML and not Latex output). I see two possibilities here:
- get blocks of code from the resulting HTML (by parsing it), highlight them and inject back,
- provide (optional) highlighting function to blackfriday which will be called during rendering parts of the code.

I believe that latter is better option both API- and performance-wise. Do you have any thoughts/opinion about this?

rtfb:
Chroma is a general purpose code highlighting library, and bfchroma
provides an easy to use glue layer for Blackfriday to interface with
Chroma.

Fixes #10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,6 +331,12 @@ are a few of note:
 *   [LaTeX output](https://bitbucket.org/ambrevar/blackfriday-latex):
     renders output as LaTeX.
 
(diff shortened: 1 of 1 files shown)


rtfb:
Thanks, @Depado! ;-)

PR #401, opened 2017-10-21
Empty link support

duzy:
Hey, russross. Thanks for your blackfriday, I'm using it for my personal website. And I came across an issue with empty links like `[](interesting metadata)`.

I'm using such links for embedding metadata in the markdown. Unfortunately, blackfriday ain't work for *empty links*, so I tweak it a bit. Please review it and check if it's useful for you and others.

Note that in this case, the `node.FirstChild` will be `nil` (which is perfect for indicating an empty link for me).


rtfb:
Please resubmit against `v2` branch. Master is left for maintenance v1 support only.

duzy:
Replaced by https://github.com/russross/blackfriday/pull/402 (against v2)

PR #402, opened 2017-10-21
Added EmptyLink support for "[](meta data goes)"

duzy:
Please see https://github.com/russross/blackfriday/pull/401 for comments.

### DIFF
--- a/inline.go
+++ b/inline.go
@@ -527,7 +527,9 @@ func link(p *Markdown, data []byte, offset int) (int, *Node) {
 
 		// links need something to click on and somewhere to go
 		if len(uLink) == 0 || (t == linkNormal && txtE <= 1) {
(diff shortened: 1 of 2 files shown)


duzy:
This fails CI, may because of `node.FirstChild == nil` (where node.Type == Link).

rtfb:
You can run the test locally with `go test -short` and debug what's causing it.

duzy:
I think I will check it the other time when available. I'm working around the other projects, so can't spend much time on this to distract myself. Besides the `node.Type == Link && node.FirstChild == nil` is very useful to me for checking *empty links* (instead of considering it as an error), and it solves my problem. Anyway it breaks the existing test suite somehow.

I will post discovered issues here if any.

Issue #411, opened 2017-11-29
Disable only smartParens


kaushalmodi:
Is there a way to disable only smartParens?

I came across that need because Blackfriday renders `(r)` in equations as `&reg;` (Registered Trademark Sign)!

So Blackfriday should either provide a way to disable SmartParens or fix this bug by making smartParens smarter by disabling it in equations.

alexaandru:
I need this as well. Would you be interested in a PR?

kaushalmodi:
I am not using blackfriday any more as Hugo switched away from this package.

PR #412, opened 2017-11-29
Add Math support (MathJaX) to blackfriday.v2

Vonng:
## Add Math support (MathJaX) to blackfriday.v2

* LaTeX math equation quoted by `$`, `$$` now can be rendered properly. Compatible with many markdown editor.  
* Markdown subscript hyphen `_` won't break subscript in LaTeX anymore.
* Enable MathJaX extension with additional Extension Flag `blackfriday.MathJaxSupport`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,3 @@ _obj
 _test*
 markdown
 tags
-
-.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ _obj
 _test*
 markdown
 tags
+
+.idea/
--- a/README.md
+++ b/README.md
@@ -184,11 +184,11 @@ implements the following extensions:
     and supply a language (to make syntax highlighting simple). Just
     mark it like this:
 
-        ```go
+        ​```go
         func getTrue() bool {
             return true
         }
-        ```
+        ​```
 
     You can use 3 or more backticks to mark the beginning of the
     block, and the same number to mark the end of the block.
@@ -209,7 +209,7 @@ implements the following extensions:
     end of the document. A footnote looks like this:
 
         This is a footnote.[^1]
-        
+
         [^1]: the footnote text.
 
 *   **Autolinking**. Blackfriday can find URLs that have not been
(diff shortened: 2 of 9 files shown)


kaushalmodi:
Thanks!

Would this also take care of this issue: https://github.com/russross/blackfriday/issues/411?

smartParens should be auto disabled inside equations so that `(r)`, `(c), etc. do not get inferred to registered trademark, copyright, etc. symbols.

Vonng:
@kaushalmodi

of course. Since smartypants render only enabled when processing `Text` nodes. It will have no effect on `inlineMath` or `blockMath` nodes.

The following example rendered properly with this PR under default configuration (+MathJaX)
```
$$
(r)(x)
$$
```

kaushalmodi:
@rtfb Can you please look at this PR?

martignoni:
This would be a big improvement as you have to disable completely Blackfriday to prevent wrong interpretations in math environments.

rtfb:
I would like to not merge any changes that significantly change the way a source document is interpreted until we have CommonMark support (meaning that we can exercise their very exhaustive set of test cases, in addition to ours).

There is another very similar PR: #288. There's some extensive discussion on it about ideas how a math extension could be implemented without altering the semantics of a source Markdown document. The gist is that MathJaX support could be added completely outside of Blackfriday core, as a drop-in renderer that does two custom things:

1. Add support for `mathjax` fenced code block language specifier, and pipe the code block through whatever processing is needed (akin to [bfchroma](https://github.com/Depado/bfchroma/) extension for code syntax-coloring).
2. In a similar fashion, override the behavior of the inline code span, and if it happens to be wrapped with `$` signs, do the same as above.

Since the above seems to be a viable idea, I'm doubly reluctant to merge this functionality into Blackfriday core without seeing it tried out (or at least discussed in more detail if others don't see it as viable as I do).

Eisfunke:
FYI: the simple $formula$/$$formula$$ syntax is used by [Pandoc](https://github.com/jgm/pandoc), as described in the manual: <http://pandoc.org/MANUAL.html#math>

entron:
Really hope this change can be merged soon!

PR #421, opened 2017-12-25
Add heading offset option to HTML renderer

emilmelnikov:
Setting HTMLRendererParameters.HeadingLevelOffset to L changes
default HTML heading levels from hX to h(X+L).


rtfb:
Thanks, @emilmelnikov!

PR #428, opened 2018-01-18
Fix Windows-style newlines

rtfb:
There still are a few XXX markers where it's clear from the code that it should deal with \r, but I couldn't pin-point a test case to exercise it yet.

Also, I'm planning to add a test runner that will translate all the `\n`s to `\r\n`s and re-run the tests.

### DIFF
--- a/block.go
+++ b/block.go
@@ -126,7 +126,7 @@ func (p *Markdown) block(data []byte) {
 		if p.isHRule(data) {
 			p.addBlock(HorizontalRule, nil)
 			var i int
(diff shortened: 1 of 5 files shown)


klingtnet:
Any plans to merge this?

rtfb:
> Any plans to merge this?

Not yet... The problem is that this PR is very much incomplete. I have a branch with much more fixes on top of this, but even they are not sufficient yet, there are a bunch of tests not passing yet.

tostercx:
Any progress on this?

Issue #336, opened 2017-03-02
Footnote markup has obsolete rel="footnote" attribute value


jturner314:
The footnote markup does not pass HTML validation (https://validator.w3.org/nu/) because `footnote` is not a valid value for the `rel` attribute. See:

* [A similar issue in Redcarpet](https://github.com/vmg/redcarpet/issues/536)
* [A list of all valid `rel` values, which does not include `footnote` (WHATWG)](https://html.spec.whatwg.org/multipage/semantics.html#linkTypes)
* [A list of obsolete `rel` values, including `footnote` (Microformats)](http://microformats.org/wiki/existing-rel-values#dropped)
* [The W3C recommendation for footnotes, which recommends just using an `a` element without `rel`](https://www.w3.org/TR/html5/common-idioms.html#footnotes)

Ideally, the `rel` attribute would be removed from footnote links.

kaushalmodi:
Bump, anyone?

PR #433, opened 2018-02-16
Fix HTML5 by removing rel="footnote" which is invalid.

IainHaslam:
As noted on #336, footnotes generate invalid HTML5 - validator.w3.org objects.

This change simply removes the (now) invalid `rel="footnote"`, and updates the tests. They all pass.

Thanks for this project - I'm using it with Hugo, so this change targets the older version. The same issue with different code is in v2. I'm happy to make a similar change there if needed.

EDIT: Travis CI builds against some go versions fail, others pass. The failures are due to breaching a 10 minute deadline, they happen on current master, and are not caused or made worse by this change.


rtfb:
LGTM. I'll merge when the restarted builds pass. Will you do the same for v2?

PR #437, opened 2018-02-25
add NoopenerLinks

jinliming2:
I think it is necessary to force rel="noopener" into the external links, even if there are other libraries such as bluemonday can do it.
The reason, blackfriday can formatting and displaying untrusted user data, but if the data is confirmed safe, it's unnecessary to sanitize data with bluemonday, and we still need to add rel="noopener" to keep site safe.
rel="noopener" is not about data, it's about **site** safe and **the style of links**. or, in other words, force rel="noopener" into the external links is the same as force target="_blank".
Why we discriminate "noopener"? Just because it is related to safe?

### DIFF
--- a/html.go
+++ b/html.go
@@ -35,6 +35,7 @@ const (
 	Safelink                                      // Only link to trusted protocols
 	NofollowLinks                                 // Only link with rel="nofollow"
 	NoreferrerLinks                               // Only link with rel="noreferrer"
+	NoopenerLinks                                 // Only link with rel="noopener"
 	HrefTargetBlank                               // Add a blank target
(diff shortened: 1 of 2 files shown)


rtfb:
I agree with the reasoning to add the `noopener`, but I wonder whether it makes sense to add a new flag for it. Maybe we could add both `noreferrer` and `noopener` when `NoreferrerLinks` is specified?

rtfb:
Yes, you're right, two attributes hidden under one name, which only implies on of them, does look confusing. I guess, that's enough reason to warrant the new `NoopenerLinks`.

PR #438, opened 2018-02-25
Don't run tests against ancient versions on Travis

rtfb:


rtfb:
@shurcooL, you crafted the last `.travis.yml`, so please take a look if I'm not inadvertently taking out something you wanted to stay.

.travis.yml
```diff
@@ -1,25 +1,9 @@
 sudo: false
 language: go
 go:
-  - 1.5.4
-  - 1.6.2
-  - tip
```
shurcooL: Is it intentional that you're removing `tip`? `tip` is the lates unreleased commit of Go. The PR description only mentions removing ancient versions.

Unless there's a reason, I think it should stay. It can help discover upcoming issues. It's an allowed failure, so it wouldn't affect the build.

However, Travis refers to it as `master` now, so if we're keeping it, update it to that.

Reference: https://docs.travis-ci.com/user/languages/go/#Specifying-a-Go-version-to-use.

.travis.yml
```diff
@@ -1,25 +1,9 @@
...
-    - go: 1.3.3
-      script:
-        - go get -t -v ./...
-        - go test -v -race ./...
-    - go: 1.4.3
-      script:
-        - go get -t -v ./...
-        - go test -v -race ./...
-  allow_failures:
-    - go: tip
```
shurcooL: This would stay too, then. Just rename `tip` to `master`.

.travis.yml
```diff
@@ -1,25 +1,9 @@
 sudo: false
 language: go
 go:
-  - 1.5.4
-  - 1.6.2
-  - tip
+  - 1.9
+  - "1.10"
```
shurcooL: I'd suggest going with `"1.9.x"` and `"1.10.x". No reason not to use latest patch version of each.

.travis.yml
rtfb: Yeah, I removed tip intentionally, should've made it clear. It was added back in 1.2/1.3 days, when certainly of Go 1 guarantee didn't yet have many years of record to prove itself. I think running tests against tip no longer serves any real purpose. But I don't feel strong about this, either way works for me. I can add it back if you think it's useful.

.travis.yml
rtfb: Yup, I didn't realize `1.9` wouldn't be doing that, I'll fix it.

.travis.yml
shurcooL: I don’t feel very strongly about it either, but I do think running against master with an allowed-failure is generally useful. It can:

- detect upcoming issues in blackfriday
- detect current issues with newer checks
- help spot regressions in new version of Go

PR #443, opened 2018-03-17
Fix consecutive lists of different kinds

mitchchn:
```
1. hello
2. world

* goodbye
* world
```

now produces two lists:

1. hello
2. world

* goodbye
* world

instead of:

1. hello
2. world
3. goodbye
4. world

### DIFF
--- a/block.go
+++ b/block.go
@@ -1291,8 +1291,24 @@ gatherlines:
 			}
 
 			// to be a nested list, it must be indented more
-			// if not, it is the next item in the same list
+			// if not, it is either a different kind of list
+			// or the next item in the same list
 			if indent <= itemIndent {
+				// are there different kinds of lists back-to-back?
+				var listTypeChanged bool
+				if p.dliPrefix(chunk) > 0 && *flags&ListTypeDefinition == 0 {
+					listTypeChanged = true
+				} else if p.oliPrefix(chunk) > 0 && *flags&ListTypeOrdered == 0 {
+					listTypeChanged = true
+				} else if p.uliPrefix(chunk) > 0 && (*flags&ListTypeOrdered != 0 || *flags&ListTypeDefinition != 0) {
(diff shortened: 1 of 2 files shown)


block.go
```diff
@@ -1291,8 +1291,24 @@ gatherlines:
 			}
 
 			// to be a nested list, it must be indented more
-			// if not, it is the next item in the same list
+			// if not, it is either a different kind of list
+			// or the next item in the same list
 			if indent <= itemIndent {
+				// are there different kinds of lists back-to-back?
+				var listTypeChanged bool
```
rtfb: Could you extract this block to a helper func?

block.go
```diff
@@ -1291,8 +1291,24 @@ gatherlines:
...
+					listTypeChanged = true
+				} else if p.oliPrefix(chunk) > 0 && *flags&ListTypeOrdered == 0 {
+					listTypeChanged = true
+				} else if p.uliPrefix(chunk) > 0 && (*flags&ListTypeOrdered != 0 || *flags&ListTypeDefinition != 0) {
+					listTypeChanged = true
+				}
+
+				if listTypeChanged {
+					*flags |= ListItemEndOfList
+					*flags &= ^ListItemContainsBlock
```
rtfb: While you're at it, could you please check if there's a nice way to avoid setting this flag on line 1290 and flipping it off again here? When reading this code, it stands out like something not directly related, would be nice to avoid if possible.

mitchchn:
Thanks for the review @rtfb. I now set the `ListItemContainsBlock` twice instead of setting it and removing it. i can't think of a way to only set it once. But it had me wondering about this behaviour:

```
1. there are 
2. no blocks

3. in this
4. list
```

Why should this list render with `<p>` tags? If we set the flag once _after_ the nesting check, it won't, but lots of tests assume this output is correct and fail.

PR #449, opened 2018-04-21
Add full info string in fenced code blocks

### LINKED ISSUE #410: Fenced code block parsing is not CommonMark compliant
blackfriday rejects opening code fences that contain more than one word, unless the words are enclosed in `{ }`.  But CommonMark explicitly _allows_ multiple words in opening fences, as can be seen at e.g. [this example in the CommonMark spec](http://spec.commonmark.org/0.28/#example-112).  (The only character *not* allowed in a code fence "[info string](http://spec.commonmark.org/0.28/#info-string)" is a backquote.)

See also [this related issue on gitea](https://github.com/go-gitea/gitea/issues/2994), and sample renderings via [Gitea](https://try.gitea.io/pjeby/jqmd/src/branch/master/jqmd.md) and [Github](https://github.com/bashup/jqmd/blob/master/jqmd.md) of a markdown document containing such multi-word opening fences.  Gitea's rendering breaks because blackfriday does not recognize them as fences, while Github's rendering uses the first word to select the language for highlighting (as shown in the example from the CommonMark spec).

As far as I can tell, the issue applies to both v1 and v2.

garfieldnate:
According to common mark, the info string for a fenced code block can be any
non-whitespace string, so adjust the code to read a full string instead of
just the syntax name.

Fixes #410 in v2.

### DIFF
--- a/block.go
+++ b/block.go
@@ -646,7 +646,7 @@ func isFenceLine(data []byte, info *string, oldmarker string) (end int, marker s
 			i++
 			i = skipChar(data, i, ' ')
 		} else {
-			for i < len(data) && !isverticalspace(data[i]){
+			for i < len(data) && !isverticalspace(data[i]) {
 				infoLength++
 				i++
 			}
--- a/block.go
+++ b/block.go
@@ -17,6 +17,7 @@ import (
 	"bytes"
 	"html"
 	"regexp"
+	"strings"
 
 	"github.com/shurcooL/sanitized_anchor_name"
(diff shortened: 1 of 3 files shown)


garfieldnate:
Let me know if anything needs to be changed. Cheers!

garfieldnate:
Ran `go fmt` to fix this one, as well.

rtfb:
Thanks a lot, @garfieldnate!

PR #453, opened 2018-04-30
Fixed HTML entity regex

9999years:
The old regex missed a lot of HTML entities, like long references (from 6-character entites like `&approx;` to the somewhat rarer `&CounterClockwiseContourIntegral;`) as well as numeric references (decimal e.g. `&#1234;` or hex e.g. `&#x13AF6;`). This fixes that.


rtfb:
Thanks a lot, @9999years!

PR #454, opened 2018-04-30
Clarified Markdown dialect is CommonMark in README

9999years:
Clarifies that Blackfriday implements CommonMark markdown; see [why is a spec needed?](http://spec.commonmark.org/0.28/#why-is-a-spec-needed-) in the CommonMark spec; either way, the dozens of different MD implementations make clarifying which spec Blackfriday uses critical.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Blackfriday [![Build Status](https://travis-ci.org/russross/blackfriday.svg?branch=master)](https://travis-ci.org/russross/blackfriday)
 ===========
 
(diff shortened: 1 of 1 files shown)


rtfb:
I'm afraid this is not right, @9999years. As of now, Blackfriday passes only around 30% of CommonMark acceptance tests. Granted, most of the failures are around whitespace tolerance and other such nitpicks, so in practice the number is somewhat higher, but still nowhere near to where it should be before we can claim conformity.

I have started some work on CommonMark support here (https://github.com/russross/blackfriday/tree/v2-commonmark-testsuite), but there's not much progress.

PR #457, opened 2018-05-25
Add rel="noopener" option

SamWhited:
For more information see https://mathiasbynens.github.io/rel-noopener/ or the Mozilla documentation: https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types


SamWhited:
Dulicate of #437, sorry, ever since GitHub messed with the default search I forget that it only searches issues now and not PRs as well.

PR #468, opened 2018-07-26
Fix html renderer escaping valid entities

### LINKED ISSUE #403: blackfriday seems to double escape valid HTML
Let's say I have the following Markdown document:

```html
<pre>quote char: &#34;</pre>
```

`&#34;` is a valid html entity number for a double quotation mark. Notably, it is the output produced by Go's <code>html.EscapeString(`"`)</code> function.

If I run it through blackfriday, the ampersand is escaped again, yielding:

```html
<p><pre>quote: &amp;#34;</pre></p>
```

It seems that blackfriday is escaping an ampersand that is part of a valid entity sequence. The equivalent would be if blackfriday turned `&amp;` into `&amp;amp;` (it does not do this).

The following sample program can demonstrate the problem. I encountered this in real code when trying to render code blocks using github.com/alecthomas/chroma, and then compiling entire markdown documents using

```
package main

import (
	"fmt"
	"html"

	blackfriday "gopkg.in/russross/blackfriday.v2"
)

func main() {
	in := html.EscapeString(`quote: "`)
	fmt.Println(string(blackfriday.Run([]byte("<pre>" + in + "</pre>"))))
}
```

Here's the Go source and rationale:

```
var htmlEscaper = strings.NewReplacer(
  	`&`, "&amp;",
  	`'`, "&#39;", // "&#39;" is shorter than "&apos;" and apos was not in HTML until HTML5.
  	`<`, "&lt;",
  	`>`, "&gt;",
  	`"`, "&#34;", // "&#34;" is shorter than "&quot;".
  )
```

https://www.w3schools.com/html/html_entities.asp

ruqqq:
This should fix #403


rtfb:
To clarify, here's what's happening (removed lines are what current test cases expect, the added ones are what @ruqqq's PR produces):

```diff
-I wish SmartyPants used named entities like `&amp;mdash;`
-instead of decimal-encoded entites like `&amp;#8212;`.
+I wish SmartyPants used named entities like `&mdash;`
+instead of decimal-encoded entites like `&#8212;`.
```

(Side note: added #469 to make clarifications such as above unnecessary)

> I can modify the test to pass by fixing the expected output. But before that, why were the expected test modelled as such?

Don't know. Looking at the history, it's been like that since forever. So my best guess is that it was a compatibility for a known shortcoming of Blackfriday (and probably Upskirt as well -- that's where the test suite was inherited from). @russross, do you have any comment on that?

If Russ has nothing to say against this, I think we're OK to modify the expected test results and go ahead with this PR.

Thanks a lot, @ruqqq!

russross:
No objection here.

I think this test was either inherited from Upskirt, or was just intended to document current behavior for regression testing. The Markdown spec left a lot undefined, so my general intention was to make sure that breaking changes were deliberate (like this) and not accidental.

moorereason:
I tested this PR against the CommonMark test suite branch (actually my PR against the branch), and this PR causes 4 tests to fail that were passing.

```
   --- FAIL: TestCommonMark/Entity_and_numeric_character_references_307 (0.00s)
       commonmark_test.go:124: Input: "&nbsp &x; &#; &#x;\n&#987654321;\n&#abcdef0;\n&ThisIsNotDefined; &hi?;\n"
           Expect: "<p>&amp;nbsp &amp;x; &amp;#; &amp;#x;\n&amp;#987654321;\n&amp;#abcdef0;\n&amp;ThisIsNotDefined; &amp;hi?;</p>\n"
           Actual: "<p>&amp;nbsp &x; &#; &#x;\n&#987654321;\n&#abcdef0;\n&ThisIsNotDefined; &amp;hi?;</p>\n"
   --- FAIL: TestCommonMark/Entity_and_numeric_character_references_309 (0.00s)
       commonmark_test.go:124: Input: "&MadeUpEntity;\n"
           Expect: "<p>&amp;MadeUpEntity;</p>\n"
           Actual: "<p>&MadeUpEntity;</p>\n"
   --- FAIL: TestCommonMark/Entity_and_numeric_character_references_314 (0.00s)
       commonmark_test.go:124: Input: "`f&ouml;&ouml;`\n"
           Expect: "<p><code>f&amp;ouml;&amp;ouml;</code></p>\n"
           Actual: "<p><code>f&ouml;&ouml;</code></p>\n"
   --- FAIL: TestCommonMark/Entity_and_numeric_character_references_315 (0.00s)
       commonmark_test.go:124: Input: "    f&ouml;f&ouml;\n"
           Expect: "<pre><code>f&amp;ouml;f&amp;ouml;\n</code></pre>\n"
           Actual: "<pre><code>f&ouml;f&ouml;\n</code></pre>\n"
```

* Test 307: "some nonentities"
* Test 309: "Strings that are not on the list of HTML5 named entities are not
recognized as entity references either"
* Tests 314 & 315: "Entity and numeric character references are treated as literal
text in code spans and code blocks"

ruqqq:
Thanks @moorereason. Will work on fixing this.

moorereason:
@ruqqq,
From [section 6.2 of the CommonMark spec](https://spec.commonmark.org/0.28/#entity-and-numeric-character-references):

> Entity references consist of `&` + any of the valid HTML5 entity names + `;`. The document https://html.spec.whatwg.org/multipage/entities.json is used as an authoritative source for the valid entity references and their corresponding code points.

Go maintains a [HTML entity map](https://github.com/golang/go/blob/master/src/html/entity.go), but it's private to the `html` package.

ruqqq:
Updated the code to pass tests 307, 309, 314 and 315 of the CommonMark suite.

moorereason:
I would recommend adding a comment to `entities.go` that acknowledges the origin of the `entities` map is the Go source (`src/html/entities.go`).  Also, put the standard blackfriday copyright comment at the top of the file.

ruqqq:
Is the copyright necessary? Some of the files does not have it (i.e. `esc.go`).

moorereason:
> Is the copyright necessary?

Well, apparently not.

jba:
Still running into this (current spec link: https://spec.commonmark.org/0.29/#example-311).

What's holding up this fix?

eparis:
@ruqqq @russross what needs to be done before this merges?

terranmoccasin:
@russross @eparis Running into this as well on a project right now. Are there plans for this PR to be merged into master?

PR #474, opened 2018-08-18
Drop race detector from Travis builds

rtfb:
We do not have a single go statement in Blackfriday code, -race does
nothing for us, but burn CPU cycles (and it hurts because we're hitting
timeouts on full test runs).


rtfb:
By zero parallelism I meant that all the logic of Blackfriday runs in a single goroutine, and there's no data that could race. But a point about speeding up tests by running them in parallel is very good one, I filed an issue for it.

Thanks!

PR #480, opened 2018-08-31
Run tests in parallel

### LINKED ISSUE #475: Run tests in parallel
Currently all tests run sequentially on a single core. We can speed things up by using [`t.Parallel()`](https://godoc.org/testing#T.Parallel).

This came up in #474, where @dmitshur suggested:

> I don't see a reason why we shouldn't be running tests in parallel by using t.Parallel(). We're currently not doing that at all. From a quick local test on a quad-core laptop, allowing tests to run in parallel speeds up go test -race gopkg.in/russross/blackfriday.v2 from 156.058s to 77.826s. I'm seeing only 200% CPU usage for most of the time (out of 800% max, given 8 threads), so I suspect there are 2 tests that take the longest to run, and can likely be sped up even further by breaking up into sub-tests that can run in parallel.

moorereason:
Fixes #475

### DIFF
--- a/block_test.go
+++ b/block_test.go
@@ -19,6 +19,7 @@ import (
 )
 
 func TestPrefixHeaderNoExtensions(t *testing.T) {
+	t.Parallel()
 	var tests = []string{
 		"# Header 1\n",
 		"<h1>Header 1</h1>\n",
@@ -88,6 +89,7 @@ func TestPrefixHeaderNoExtensions(t *testing.T) {
 }
 
 func TestPrefixHeaderSpaceExtension(t *testing.T) {
+	t.Parallel()
 	var tests = []string{
 		"# Header 1\n",
 		"<h1>Header 1</h1>\n",
@@ -148,6 +150,7 @@ func TestPrefixHeaderSpaceExtension(t *testing.T) {
 }
 
 func TestPrefixHeaderIdExtension(t *testing.T) {
+	t.Parallel()
 	var tests = []string{
 		"# Header 1 {#someid}\n",
 		"<h1 id=\"someid\">Header 1</h1>\n",
@@ -208,6 +211,7 @@ func TestPrefixHeaderIdExtension(t *testing.T) {
 }
 
 func TestPrefixHeaderIdExtensionWithPrefixAndSuffix(t *testing.T) {
+	t.Parallel()
 	var tests = []string{
 		"# header 1 {#someid}\n",
 		"<h1 id=\"PRE:someid:POST\">header 1</h1>\n",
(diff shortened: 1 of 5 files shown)


rtfb:
Would it work if we move all these calls to `t.Parallel` to the `doTests*` functions? That would significantly reduce repetition.

moorereason:
@rtfb,
That's the first thing I tried, but it didn't enable parallelism.  I'm not 100% sure why, but it didn't work (and I just tried again to confirm).  🤷‍♂️

rtfb:
Yeah... I checked the documentation for the clues before asking, and there wasn't anything indicating why it would't, so well, go figure. Let me try to dig for a bit, and we can merge it as is if nothing turns up -- always can fix it up later.

rtfb:
Well, I just tried this:

```diff
diff --git a/helpers_test.go b/helpers_test.go
index 089c730..b9cf466 100644
--- a/helpers_test.go
+++ b/helpers_test.go
@@ -69,6 +69,7 @@ func doTestsBlock(t *testing.T, tests []string, extensions Extensions) {
 }
 
 func doTestsParam(t *testing.T, tests []string, params TestParams) {
+       t.Parallel()
        execRecoverableTestSuite(t, tests, params, func(candidate *string) {
                for i := 0; i+1 < len(tests); i += 2 {
                        input := tests[i]
@@ -147,6 +148,7 @@ func transformLinks(tests []string, prefix string) []string {
 }
 
 func doTestsReference(t *testing.T, files []string, flag Extensions) {
+       t.Parallel()
        params := TestParams{extensions: flag}
        execRecoverableTestSuite(t, files, params, func(candidate *string) {
                for _, basename := range files {
```

and it did run in parallel. Looking at `go test -v`, the output has this pattern:

```
=== RUN   TestReference_EXTENSION_NO_EMPTY_LINE_BEFORE_BLOCK
=== PAUSE TestReference_EXTENSION_NO_EMPTY_LINE_BEFORE_BLOCK
=== CONT  TestPrefixHeaderNoExtensions
--- PASS: TestPrefixHeaderNoExtensions (0.00s)
```

mostly consisting of `RUN+PAUSE` at first (I see that as a sign of queueing up the tests), and mostly of `CONT+PASS` closer to the end. Also, watching `top` indicated a CPU utilization slightly above 100% before the changes, and well above 200% after.

The problem, however, was that these were not the only changes I had to make to get that working... There's a handful of tests (maybe approaching a dozen) that do multiple calls to the `doTests*` helpers. I commented them out for quick check, but the right thing to do is to hunt them down and convert them into subtests via `t.Run()`.

moorereason:
Hmm.  I'll take another look.

dmitshur:
@rtfb FWIW, I think calling `t.Parallel()` at the top of every test, as done in this PR at this time, is a good way of marking tests as parallel.

It is more verbose, but the benefit is that it makes it more explicit, easy to see if a test is parallel, and easier to avoid panic due to `t.Parallel()` being called more than once (e.g., if one helper happens to use another, and both call `t.Parallel()`.

moorereason:
I agree with @dmitshur.  I'm ready to merge.  :-)

rtfb:
I couldn't say I'm convinced that this argument applies to our specific case, but I could say that this is too minor an issue to quarrel about, so let's just merge it in :-)

PR #482, opened 2018-08-31
Refactor CommonMark tests to be Go tests

moorereason:
This commit refactors the CommonMark tests as actual Go tests. The tests
use Go sub-tests for each CommonMark section, so you can now select a
subset of tests (using `-run=`).

I tried to maintain the report features, but they are printed above the
FAIL notices (a small incentive to fix some tests). The report now only
shows sections that were tested if you selected a subset of tests.

The tests are protected by a build tag now, so this branch could be
merged upstream without breaking anything.

The JSON file has also been updated to the latest from the CommonMark
project.


moorereason:
Closing due to lack of interest.

PR #490, opened 2018-09-17
Add PostRefOverride functionality

### LINKED ISSUE #467: feature request: Ref override after references resolved
Currently the [`WithRefOverride`] option is processed before references are resolved. This makes good sense in some cases, eg. I am currently using it to resolve DOI URI's (eg. `doi:1234`) to HTTP links. However, in other cases it may be that we don't want to override links already set by the user. For example, consider the following:

```
Look at [RFC 7622] [pdf] or [XEP-0198].

[RFC 7622]: https://tools.ietf.org/pdf/rfc7622.pdf
```

In this example I want to automatically link to RFCs and XEPs so that the user can reference them without providing a link. The XEP link would be generated by my ref override, but what if they already have an RFC link and I override the PDF to the HTML version?

It would be useful to have a ref override feature that occurs after refs have already been processed to make this possible. This way I wouldn't override the users ref for RFC7622, but could still automatically genreate a ref for other RFCs.

[`WithRefOverride`]: https://godoc.org/gopkg.in/russross/blackfriday.v2#WithRefOverride

SamWhited:
I mimicked the `WithRefOverride` option for this functionality, and hope that it or something like it will be accepted. I would like to be able to override refs but only if the user has not specified a reference, which means that ref overriding needs to happen after the refs have been resolved.

Fixes #467

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,1 +1,6 @@
 module github.com/russross/blackfriday/v2
+
+require (
+	github.com/pmezard/go-difflib v1.0.0
+	github.com/shurcooL/sanitized_anchor_name v0.0.0-20170918181015-86672fcb3f95
+)
--- a/go.sum
+++ b/go.sum
@@ -1,0 +1,4 @@
+github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
+github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
+github.com/shurcooL/sanitized_anchor_name v0.0.0-20170918181015-86672fcb3f95 h1:/vdW8Cb7EXrkqWGufVMES1OH2sU9gKVb2n9/1y5NMBY=
+github.com/shurcooL/sanitized_anchor_name v0.0.0-20170918181015-86672fcb3f95/go.mod h1:1NzhyTcUVG4SuEtjjoZeVRXNmyL/1OwPU0+IJeTBvfc=
--- a/helpers_test.go
+++ b/helpers_test.go
@@ -25,7 +25,6 @@ import (
 type TestParams struct {
 	extensions        Extensions
 	referenceOverride ReferenceOverrideFunc
-	postRefOverride   PostReferenceOverrideFunc
 	HTMLFlags
 	HTMLRendererParameters
 }
@@ -51,8 +50,7 @@ func runMarkdown(input string, params TestParams) string {
 	renderer := NewHTMLRenderer(params.HTMLRendererParameters)
 	return string(Run([]byte(input), WithRenderer(renderer),
 		WithExtensions(params.extensions),
-		WithRefOverride(params.referenceOverride),
-		WithPostRefOverride(params.postRefOverride)))
+		WithRefOverride(params.referenceOverride)))
 }
 
 // doTests runs full document tests using MarkdownCommon configuration.
--- a/helpers_test.go
+++ b/helpers_test.go
@@ -25,6 +25,7 @@ import (
 type TestParams struct {
 	extensions        Extensions
 	referenceOverride ReferenceOverrideFunc
+	postRefOverride   PostReferenceOverrideFunc
 	HTMLFlags
 	HTMLRendererParameters
 }
@@ -50,7 +51,8 @@ func runMarkdown(input string, params TestParams) string {
 	renderer := NewHTMLRenderer(params.HTMLRendererParameters)
 	return string(Run([]byte(input), WithRenderer(renderer),
 		WithExtensions(params.extensions),
-		WithRefOverride(params.referenceOverride)))
+		WithRefOverride(params.referenceOverride),
+		WithPostRefOverride(params.postRefOverride)))
 }
 
 // doTests runs full document tests using MarkdownCommon configuration.
--- a/html.go
+++ b/html.go
@@ -217,6 +217,10 @@ func skipSpace(tag []byte, i int) int {
 }
 
 func isRelativeLink(link []byte) (yes bool) {
+	if len(link) == 0 {
+		return false
+	}
+
 	// a tag begin with '#'
 	if link[0] == '#' {
 		return true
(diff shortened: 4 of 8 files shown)


rtfb:
Hi Sam, thanks for taking a stab at it!

I'm trying to wrap my head around this feature, and all the aspects of it... Wouldn't an existing reference override functionality also solve that if we just moved it in place of `postRefOverride`? Provided that we extend the callback to be able to receive params?

It also seems that your implementation can't really handle what you described in your feature request. If I'm reading it correctly, a non-nil `p.referenceOverride` *will* get executed and will override your reference, robbing `postRefOverride` of the chance to get executed. Is this right?

Looking broader, I'm wondering what could be a better way to upgrade the ref override feature to v2 world. It's a relic from v1 times, where there was no better way to override references. In v2, much more elegant and flexible solution should be possible.

The `getRef` function is only called in two places, both in inline parser's link resolver. What if, instead of calling `getRef` and juggling with callbacks there, we just extend the `LinkData` struct to have everything that's needed to override/postprocess the reference links as a separate step after parsing? Thinking minimalistically, every `LinkData` could contain these additional fields:

```go
type LinkData struct {
    // ...
    IsRefLink bool
    RefID string
    Refs *map[string]*reference    // a pointer to what's now p.refs
    WasOverridden bool
}
```

Right after the parsing stage, `IsRefLink` would be set to `true`, `RefID` to what is now passed to `getRef`, `Refs` to `&p.refs`, and `WasOverridden` to `false`. A renderer would take that quadruplet and simply look up `node.Refs[node.RefID]` (effectively doing an equivalent of what `getRef` does today sans calling `p.referenceOverride`).

But, if the caller had chosen to traverse the tree and do custom stuff on those nodes, she could set `WasOverridden` to `true`, and renderer would know to render that node verbatim, without looking things up once more.

Does that make sense? Would you be interested in giving it a go?

SamWhited:
> Wouldn't an existing reference override functionality also solve that if we just moved it in place of postRefOverride

I would be all for this if it's possible and breaking the API isn't something you're worried about. Reading the rest of your reply you mentioned things happening in a "v2" world: does this mean that v2 isn't stable yet or recommended for use? I was under the impression that it was the latest thing, but gopkg.in and github were confusing me.

> It also seems that your implementation can't really handle what you described in your feature request. If I'm reading it correctly, a non-nil p.referenceOverride will get executed and will override your reference, robbing postRefOverride of the chance to get executed. Is this right?

I'll have to play with it; I tried it and wrote some tests and it *seemed* to be working, but I don't fully understand the reference parsing or the difference between different kinds of references.

> …

RE the rest of your message: I'll happily take a stab at it pending an answer to the first part of my reply about whether v2 is recommended and/or not a stable API yet. Thanks for all the help and the information (and all the work you do on this library)! I'll have to dig through and make sure I understand everything better.

SamWhited:
> The getRef function is only called in two places, both in inline parser's link resolver. What if, instead of calling getRef and juggling with callbacks there, we just extend the LinkData struct to have everything that's needed to override/postprocess the reference links as a separate step after parsing?

I started experimenting with this today. Adding `IsRefLink` and `RefID` was easy enough, but right now the parser ignores links with no matching reference and just renders them as text blocks. I'm not sure how best to handle that yet.

> Refs *map[string]*reference

What would the use case be for storing the entire reference map? Shouldn't it store instead the single reference returned by `getRef` for the particular link?

rtfb:
> the parser ignores links with no matching reference and just renders them as text nodes. I'm not sure how best to handle that yet.

I think the parser tries doing too much now: it tries to make decisions about the reference while what it's ought to do is just saying "ok, I've found a reference syntax here, and the syntax was valid, and here are the details about it". And then the renderer (or whoever else might consume the AST) can decide what to do about it: render a valid reference if it's found; render a broken link if not found; render a broken reference as plain text; render a broken reference as a link to a catch-all 404 page, etc.

In other words, we should not have logic like this:

```go
		lr, ok := p.getRef(string(id))
		if !ok {
			return 0, nil
		}
```

it's not a parser's job to decide what to do in case of `!ok`, it should just keep recording everything it needs to know to construct a `Node`.

> What would the use case be for storing the entire reference map? Shouldn't it store instead the single reference returned by getRef for the particular link?

So that we wouldn't need to expose it in any other way. Currently, `Markdown.Parse` returns a single thing: a pointer to the root node of the AST. Having the link nodes to contain a pointer to the entire reference map was the easiest way I could think of that would preserve the simplicity of this interface. No additional return values, no additional structs defined. (But I might have missed something! If you have other suggestions, let's discuss!)

SamWhited:
> In other words, we should not have logic like this:

Excellent; I'd stripped those out to make it work. I'll move that logic into the renderer to make tests pass again.

SamWhited:
Actually, figuring out the type of link and reconstructing markdown in the renderer feels poor too. I'll have to play around with it some more.

SamWhited:
I've pushed a start to this; tests are mostly failing still. I'll keep trying to figure out the renderer part later, but I probably don't have a good enough grasp of how all the pieces fit together yet.

SamWhited:
Tests like the following:

```
[Something in brackets like [this][] should work]
```

appear to be the problem. If we just try to treat everything like a reference, the outer reference is the only thing returned in this case.

rtfb:
This, `[Something in brackets like [this][] should work]`, should not be a problem, because this bit `[this][]` is a complete and unambiguous reference style link syntax, so we know we can pick it up.

However, I didn't realise this shorthand syntax is also valid: `[Same with [this].]`. Can we resolve that by only treating the inner-most nested brackets as a reference? At least on the surface, it seems like we should.

SamWhited:
Sounds good; this appears to be the behavior CommonMark uses, though I have no idea if that's specified or an implementation detail. Either way, multiple things seem to do it so it sounds good to me. I'll see if I can figure it out.

SamWhited:
I don't think I'm going to get back to this any time soon; sorry. I just don't understand the parser well enough to get this done in a reasonable timeframe, and I've been trying to cut back on my external contributions. Closing for now, and I'll reopen if I have the time to get back to it. Thanks for all your help with this and for your excellent library!

rtfb:
> I don't think I'm going to get back to this any time soon; sorry. I just don't understand the parser well enough to get this done in a reasonable timeframe, and I've been trying to cut back on my external contributions. Closing for now, and I'll reopen if I have the time to get back to it. Thanks for all your help with this and for your excellent library!

No worries. That code is indeed quite complicated and involving, so probably a bit steep for first acquaintance with the codebase. The very fact that we got this idea fleshed out and documented is quite valuable in itself. Thanks!

PR #492, opened 2018-09-26
Add more info to LinkData

SamWhited:
I've been having trouble keeping track of how each thing in #490 affects every other thing. To keep things simple, if having more link info would be desirable either way I thought I'd split this out into a separate PR that can be merged before the other one is ready. I'll rebase 490 on top of this branch, so even if you don't want this yet we can close the PR and the commits will be separate at least.


rtfb:
Yup, this looks like a good direction, but let's start merging things when the bigger picture gets in sight.

PR #502, opened 2018-11-07
Run go mod tidy

SamWhited:
Run `go mod tidy` to pull in all dependencies so that building won't keep writing changes to the `go.mod` file that have to be skipped over when adding files to version control.


rtfb:
Thanks, Sam!

Issue #348, opened 2017-04-13
Things to forward port from v1 to v2


rtfb:
A tracking issue to mark stuff that needs forward porting from v1 to v2.

API-breaking (i.e. must be addressed before releasing v2):
* _No known issues_

Other:
* #334

james-lawrence:
looks like this is done?

dmitshur:
This isn't done; those are three v1 PRs that need to be ported to v2. I'll work on #352 now.

PR #509, opened 2018-12-23
Document SanitizedAnchorName algorithm, copy implementation.

dmitshur:
The goal of this change is to reduce number of non-standard library packages (repositories) that blackfriday imports (not counting imports used only for tests) from 1 to 0, and in turn, reduce the cost of importing blackfriday into other projects.

Do so by documenting the algorithm of `SanitizedAnchorName`, and include a copy of the small function inside blackfriday itself. The same functionality continues to be available in the original location, `github.com/shurcooL/sanitized_anchor_name.Create`. It can be used by existing users and those that look for a small package, and don't need all of blackfriday functionality. Existing users of blackfriday can use the new `SanitizedAnchorName` function directly and avoid an extra package import.

This change is a port of PR #352 from v1 into v2.

Updates #348.
Updates #350.


dmitshur:
Friendly ping @rtfb, do you have a chance to review this?

/cc @julieqiu I think this PR may fix the issue that you ran into.

rtfb:
Yep, looks good. Thanks, and sorry for the delay.

dmitshur:
No problem! Thanks, I'll merge it now.

rtfb:
Yeah, makes sense. I also want to check how https://github.com/russross/blackfriday/pull/515 works with broken reference tests, and if it pulls it off, we could merge that one as well. I'll try to do that in the coming days.

PR #511, opened 2018-12-23
Fix checking of backslash in link text (#510)

### LINKED ISSUE #510: double backslashes before `]` break link
When the end of the link text is a double backslash, the link breaks.

E.g:
```
[link\\](/url/)
```

It should be as follows:

```
<p><a href="/url/">link\</a>.</p>
```

But got:
```
<p>[link\](/url/).</p>
```

This behavior was confirmed in both v1 and v2.

autopp:
Fix #510.

By using `isBackslashEscaped`, it prevents incorrect interpretation of backslash in link.

### DIFF
--- a/inline.go
+++ b/inline.go
@@ -278,7 +278,7 @@ func link(p *Markdown, data []byte, offset int) (int, *Node) {
 		case data[i] == '\n':
 			textHasNl = true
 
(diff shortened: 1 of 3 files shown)


rtfb:
Thank you, @autopp!

PR #519, opened 2019-01-08
readme: use fenced code blocks

Quasilyte:
It's more consistent and it's possible to mark code blocks with "go" hint,
so we get syntax highlighting on GitHub.


rtfb:
Yup, makes sense. Thank you, @Quasilyte!

Issue #495, opened 2018-10-08
Block parsing for fenced code blocks with syntax highlighting in lists regression.


dmitshur:
I've just discovered PR #476 is causing the following regression in v1. /cc @tfogo @client9

Given this Markdown input:

<pre>
### Codeblock within list

-	list entry 1

	```
	if (i == 5)
	    break;
	```

-	list entry 2

### Stuff that comes later

stuff here
</pre>

Everything works as expected:

![image](https://user-images.githubusercontent.com/1924134/46589824-3e03ec00-ca7c-11e8-9ef9-e7510d013692.png)

However, if one uses a fenced code block with syntax highlighting, everything that comes after the fenced code block ends up being a part of "list entry 1":

<pre>
### Codeblock within list

-	list entry 1

	```C
	if (i == 5)
	    break;
	```

-	list entry 2

### Stuff that comes later

stuff here
</pre>

![image](https://user-images.githubusercontent.com/1924134/46589839-57a53380-ca7c-11e8-86aa-26145208ff57.png)

I.e., it parses _as if_ the input was:

<pre>
### Codeblock within list

-	list entry 1

	```C
	if (i == 5)
	    break;
	```

	-	list entry 2

	### Stuff that comes later

	stuff here
</pre>

dmitshur:
I have a lead. Modifying the code as follows seems to fix the issue, at least for that one instance of the problematic input:

```diff
-_, marker := isFenceLine(chunk, nil, codeBlockMarker, false)
+_, marker := isFenceLine(chunk, new(string), codeBlockMarker, false)
```

I.e., passing a non-nil `info *string` parameter to `isFenceLine`.

SamWhited:
I came here to report a similar issue which sounds like it might have the same cause, right now if a fenced code block immediately follows a list even if it's not indented it becomes part of the list. The following patch adds a test that can be used to demonstrate the behavior (the expected result might have wrong newlines, I just guessed):

```
diff --git a/block_test.go b/block_test.go
index 326c3110ce70..4db7d24156e2 100644
--- a/block_test.go
+++ b/block_test.go
@@ -1135,6 +1135,9 @@ func TestFencedCodeBlock(t *testing.T) {

                "```\n[]:()\n[]:)\n[]:(\n[]:x\n[]:testing\n[:testing\n\n[]:\nlinebreak\n[]()\n\n[]:\n[]()\n```",
[...]

PR #521, opened 2019-01-18
RFC Attempt to fix fenced code in lists issue

### LINKED ISSUE #485: Regression: a list is broken if an item starts with a fenced code block
Input:
```markdown
-   ```
    foo
    ```

-   bar
```

Expected output (the first item has `<pre>`):
```html
<ul>
<li><pre><code>foo
</code></pre></li>

<li><p>bar</p></li>
</ul>
```

New output since f1f45ab (the first item lacks `<pre>`; the second item is not recognized as an item):
```html
<ul>
<li><code>
foo
</code>

-   bar</li>
</ul>
```

aignas:
This attempts to fix #485 and #495.

Note the test cases which were added at the bottom of the list. The first added test case was passing even before the changes, but the second was not.

Comments are welcome if the change in the existing behaviour is welcome.

### DIFF
--- a/block.go
+++ b/block.go
@@ -649,8 +649,10 @@ func isFenceLine(data []byte, info *string, oldmarker string, newlineOptional bo
 	}
 
 	i = skipChar(data, i, ' ')
-	if i >= len(data) || data[i] != '\n' {
+	if l := len(data); i >= l || data[i] != '\n' {
 		if newlineOptional && i == len(data) {
+			return i, marker
+		} else if !newlineOptional && i < l && data[l-1] == '\n' {
 			return i, marker
 		}
 		return 0, ""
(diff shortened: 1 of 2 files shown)


block.go
```diff
@@ -649,9 +649,11 @@ func isFenceLine(data []byte, info *string, oldmarker string, newlineOptional bo
 	}
 
 	i = skipChar(data, i, ' ')
-	if i >= len(data) || data[i] != '\n' {
+	if l := len(data); i >= l || data[i] != '\n' {
```
rtfb: Somewhat minor nit: could you please change the name of `l` to something else? It's too similar to `i` and `1` and is difficult to read.

Also, this entire chain of conditionals is becoming very convoluted. A pattern like

```go
if cond ... {
  return X
} else if !cond ... {
  also return X
}
```

is a sure sign this should be simplifiable. I'll try to get back with a more concrete suggestion, but just leaving this note here for now.

rtfb:
In it goes. Thanks a lot, @aignas!

aignas:
@rtfb, thanks for taking time to review it!

PR #530, opened 2019-03-07
v2 block: remove fencedCodeBlock consumption inside blockQuotes

bobertlo:
See https://github.com/bobertlo/vmd/issues/7

This was giving me some weird issues, and doesn't seem to break anything by removing it. Not sure what the purpose was.


bobertlo:
Oh, I see now that it breaks the HTML. I'm only using the AST which is pretty buggy!

I'll maybe try to see if I can figure out a way for both to work, otherwise I'm vendoring off of this fork anyways so I'm good.

PR #540, opened 2019-04-23
Export Node.IsContainer and add Node.IsLeaf

twpayne:
These functions are very useful for applications that walk the AST generated by `blackfriday`. Without these exports, the applications have to work out which are leaf nodes and which are container nodes themselves.

### DIFF
--- a/node.go
+++ b/node.go
@@ -199,7 +199,8 @@ func (n *Node) InsertBefore(sibling *Node) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


twpayne:
If you prefer, I could make `IsContainer`/`IsLeaf` methods on `NodeType` instead.

twpayne:
@russross Are there any changes that you'd like to this PR before it can be merged?

PR #547, opened 2019-06-10
Use ↩ as default params.FootnoteReturnLinkContents

adiabatic:
This changes the default `params.FootnoteReturnLinkContents` from `<sup>[return]</sup>` to `<span aria-label='Return'>↩\ufe0e</span>`.

It’s very common to use `↩` in footnote links. However, some platforms like iOS and iPadOS choose to use emoji presentation for this particular character. This leads to lots of blogs, by default, looking silly on portable Apple gizmos, as described in <https://github.com/jgm/pandoc/issues/5469>. By switching to a return arrow with a disable-emojification variation selector, we get blackfriday to do the right thing by default.

Additionally, ↩ is more language-agnostic than “return” is, so blackfriday will work better out of the box for more people.

### DIFF
--- a/html.go
+++ b/html.go
@@ -132,7 +132,10 @@ func NewHTMLRenderer(params HTMLRendererParameters) *HTMLRenderer {
 	}
 
 	if params.FootnoteReturnLinkContents == "" {
-		params.FootnoteReturnLinkContents = `<sup>[return]</sup>`
+		// U+FE0E is VARIATION SELECTOR-15.
+		// It suppresses automatic emoji presentation of the preceding
+		// U+21A9 LEFTWARDS ARROW WITH HOOK on iOS and iPadOS.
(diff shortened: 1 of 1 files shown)


dmitshur:
The documentation for the [`HTMLRendererParameters.FootnoteReturnLinkContents`](https://pkg.go.dev/github.com/russross/blackfriday/v2#HTMLRendererParameters.FootnoteReturnLinkContents) field also needs to be updated. It currently describes the old default value:

```
$ go doc github.com/russross/blackfriday/v2 HTMLRendererParameters.FootnoteReturnLinkContents
package blackfriday // import "github.com/russross/blackfriday/v2"

type HTMLRendererParameters struct {
    // Show this text inside the <a> tag for a footnote return link, if the
    // HTML_FOOTNOTE_RETURN_LINKS flag is enabled. If blank, the string
    // <sup>[return]</sup> is used.
    FootnoteReturnLinkContents string

    // ... other fields elided ...
}
```

PR #578, opened 2019-12-04
Implement a furigana extension

Elinvention:
Implement a furigana extension based on the syntax used in https://github.com/djfun/furigana_markdown. See https://discourse.gohugo.io/t/using-furigana-ruby-with-markdown/15156/4 to understand what this is for.

I made this some months ago and now I realized Hugo switched to goldmark, oh well...


russross:
This is probably too niche to merge in the base library. I think it would make more sense to maintain this in a fork.

PR #584, opened 2020-03-13
Fix bug: AutoHeadingIDs being ignored

AlmogBaku:
AutoHeadingIDs being ignored


Fornax96:
Can this be merged? Very annoying that writeTOC overwrites the header IDs. This makes it very hard to generate clean readable URLs for my documentation pages.

PR #586, opened 2020-03-24
go.mod: add go version

kolyshkin:
This is a result of running `go mod tidy` using go 1.13.6.

In my devel toolchain (vim-go -> gopls -> go 1.13), go apparently
wants to add this line to go.mod, but since the modules are read-only,
it fails, which leads to gopls failing, too.

Merging this and tagging a release would help tremendously
(not because of this PR per se, but mostly because of
https://github.com/russross/blackfriday/pull/509 and https://github.com/russross/blackfriday/pull/515 
which are not in a released version yet).

@russross @rtfb @SamWhited PTAL

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,1 +1,3 @@
 module github.com/russross/blackfriday
+
+go 1.13


SamWhited:
Is it trying to modify a `go.mod` file in your cache? I don't understand why it would be readonly otherwise. If so, that sounds like a bug in gopls or Go itself to me.

That being said, this LGTM.

kolyshkin:
> Is it trying to modify a go.mod file in your cache?

Yes, in is in the cache, sorry I wasn't clear earlier.

I'm not sure if it's a fatal error or just a warning, but it shows an error

> vim-go: error writing go.mod: open /home/kir/go/pkg/mod/github.com/russross/blackfriday/v2@v2.0.1/go.mod298498081.tmp: permission denied

(in here vim-go is just a prefix added by vim-go, the actual error I guess comes from gopls but I'm not 100% sure)

If I modify `go.mod` file the same way as in this PR this error is gone.

> If so, that sounds like a bug in gopls or Go itself to me.

Probably. I am going to open a bug for gopls once I dig deeper.

kolyshkin:
Update: the error seems to be caused by `guru` tool (not `gopls` as I was assuming).

Issue #587, opened 2020-03-24
v2: decide on import path and make a new release


kolyshkin:
The latest release is from September 2018, maybe it's time to make a new one?

The primary reasons for a new release are #509 and #515; this will make the package not depend on any other packages. Including #586 would be nice, too.

dmitshur:
This should be done.

In order to be able to publish new v2 versions, a blocking task is to resolve the current lack of clarity about what import path should be suggested and continued to be used for v2: `github.com/russross/blackfriday/v2` or `gopkg.in/russross/blackfriday.v2`. The `go.mod` file says one thing, the README on master branch says something that isn't very consistent.

By making a newer release, some decision would be committed to, and ideally that decision should be a good one given the constraints this project is currently operating with. Some of the decisions made in the past predate a good understanding of modules and best practices regarding v2 versions, which has made this decision more difficult to make conclusively.

I suspect that in the current situation, the best course of action is to continue to use the `github.com/russross/blackfriday/v2` import path, as defined by the module statement in go.mod file in v2 branch:

https://github.com/russross/blackfriday/blob/3b2b20c8cb0632a0442ea9694d170b8aea985549/go.mod#L1

I started talking to @rtfb about resolving this situation a number of months ago, but then unfortunately dropped the ball on it due to lack of bandwidth. But I wanted to post this comment to shed light on what I believe is a large contributing factor to no new v2 releases.

russross:
I'm happy with either import path and agree that we should post a new release soon. I haven't been following developments in the module system very closely--is there one approach that is more forward looking/will fit in better with the community than the other?

SamWhited:
There's also a third option which I always personally prefer: using a new vanity import path.

This requires that someone own a domain name and setup an HTTP server (or use a service like Netlify or GitHub Pages) which is more work for the maintainers, however, it gives you the benefit of owning your package import instead of relying on GitHub or gopkg.in never changing or going under or being changed to `github.microsoft.com` or similar. While this may not seem likely, other scenarios could be imagined where GitHub becomes impractical in the future, for example a price change, adding more extreme rate limiting that causes fetching packages to break more often, the author just decides they don't like GitHub and want to use <trendy new future service>, etc.
Using a vanity import keeps things flexible in the future, and avoids unforeseen future issues at the cost of a tiny bit of infrastructure work. I would be willing to help set this up if you want.

kolyshkin:
> This requires that someone own a domain name and setup an HTTP server

[...]

PR #676, opened 2020-11-26
Fix 404 error for v2

akumbhani66:


dmitshur:
The `github.com/russross/blackfriday/v2` import path is intentional. See https://github.com/russross/blackfriday/issues/587#issuecomment-703393820.

Where are you getting a 404?

dmitshur:
Closing due to no answer from PR author. We can re-open if there's new information.

PR #679, opened 2021-02-03
v2: handle parentheses inside of links

kovetskiy:
This is basically the same as #222 but for the v2 branch.

All tests pass.


kovetskiy:
Unfortunately, I have to close the PR because I need to be able to use these changes without using the `replace` directive in the `go.mod` file. But I'd be glad to re-open PR if Russ (or other maintainers) will have time for the project again.

PR #685, opened 2021-06-15
Ignore underscore conditionally with the IgnoreUnderscore extension

dmitrytorba:
This PR adds an extension `IgnoreUnderscore` that can be used to skip processing all underscores. Similar to `NoIntraEmphasis` this extension can be used by users to control the rendering behavior of underscores.

BTW, thanks for the project! Its a very useful tool.


enahs:
@dmitshur can you please review this PR 🙏

PR #710, opened 2023-09-08
fix cell alignment

Skeeve:
According to https://www.w3schools.com/Css/css_table_align.asp and to my experience of using https://github.com/kovetskiy/mark, alignment of cells with blackfriday does not work correctly.

This patch fixes the issue by inserting `style="text-align: <alignment>"` into each `td` and `th`.


Skeeve:
Update. According to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/td the align-property in cells is deprecated.

Skeeve:
Seems no one is interested and maybe it doesn't even apply to xhtml.

